PR #30, opened 2020-02-05
Episode data

vinault:


BDonnot:
@vinault I htink there are some troubles in the "Observation.py" file:
```python
import json
import os
<<<<<<< HEAD
import math
=======
import time
>>>>>>> 526c019ac6f47c40707ac5b0beefc26cff61ef30
```
I tried to fix it, but unfortunately the tests made for the runner are not passing.

The lack of documentation doesn't help investigating, but i have the impression the runner computes one more time step.

Also, the name "indx" used to initialized the EpisodeData is confusing. It should be a string, and renamed something like "episode_name". Having a method of chronic_handler that returns this name is probably a good idea.
The "get_dataframes" variable used in makeing the EpisodeData is also not used.
The funciton "to_disk" of this same class EpisodeData doesn't use the names for "episode_times" (EpisodeData.TIMES), "_parameters.json" (EpisodeData.PARAMS) or "episode_meta.json" (EpisodeData.META) as defined in the class.

I would also have appreciated a test in test_EpisodeData that ensures me that running the episode does something, but when you load it back, you get the same results.
Also, what happens when you run the runner in parallel ? It's not clear in this setting.

Issue #42, opened 2020-03-16
Numba Req on Ubuntu


zhampel:
Fresh command line install of Grid2Op on Ubuntu 18.04 resulted in `numba` package not being found. Following error occurred when running `python -m grid2op.main` with python-3.6.9:

```
numba cannot be imported and numba functions are disabled.
Probably the execution is slow.
Please install numba to gain a massive speedup.
(or if you prefer slow execution, set the flag numba=False to avoid this warning!)
```

Will include `numba` to install_requires in setup.py and submit PR.

BDonnot:
Hello,

Thanks for this fix :-)
I handled the problem a bit differently. I didn't want to prevent the installation of grid2op on systems on which `numba` can't be installed. That is why i didn't add it as a dependency.

 I made two different changes in the latest grid2op version, to improve the situation regarding numba!:

- if `numba` can be installed, the installation script will try to install it, and "pass" if it can't (thus ensuring everyone can still install it even without numba)
-  The warning thrown when numba is not installed is now silenced. Only one warnings appears when first loading the "Environment" module if numba is not installed. This should reduce the "spamming" of the message.

If you think these two modifications are "ok" don't hesitate to close the issue. Otherwise i'm happy to further discuss what can be done to handle the situation even better

Thanks

Benjamin

zhampel:
Sounds good Benjamin! Thanks for making clear your changes as well.

PR #43, opened 2020-03-16
Adds numba to install req line in setup script

zhampel:
Adds a call to install numba in install requirements line in `setup.py`. Addresses issue #42


BDonnot:
I close this issue as i implemented something similar in a different way, as better described in the issue #42 .

Thanks for this proposition

Benjamin

PR #91, opened 2020-05-19
Antoine modi

marota:
Added a new mainteance generator in grid2op/Chronics/GridStateFromFileWithForecasts.py, by defining a child class GridStateFromFileWithForecastsWithMaintenance


BDonnot:
Development has to be done in bdonnot and not in RTE :-)

PR #95, opened 2020-05-20
EpisodeReplay gif edge case

Tezirg:
Dont try to write a gif if there are no frames.


BDonnot:
I'll finally update a new grid2op version

PR #161, opened 2020-11-04
Fixes notebooks and update the environment educ_case14_redisp

BDonnot:
fixing notebooks and adding real data for the educ_case14


BDonnot (commented):
This looks good

PR #167, opened 2021-01-07
Mj devs pr

mjothy:
Added EpisodeData to return API of runs in runner.
This required some changes to EpisodeData class, which was only storing in memory some data if the disk serialisation flag was on. I added a new flag for this in memory storing to keep it optional since it can take a lot of space.
Previous behavior is still available through old API, I just added a new method 'run_detailed' that enables the additional data.


BDonnot:
I think this PR has been merged in the dev repo (bdonnot/grid2op). If that is not the case, then let me know :-)

Benjamin

Issue #180, opened 2021-04-06
In "config.py" have the thermal limit set as a dictionnary


BDonnot:
## Is your feature request related to a problem? Please describe.

Today, when you set the thermal limit, you give an ordered list, this is not really convenient. Make the possibility to give them as a dictionnary with keys: line names, values: thermal limit (in A)

PR #183, opened 2021-04-07
Bd dev

BDonnot:
- [FIXED] Issue #181 : now environment can be created with  a layout and a warning is issued in this case.
- [FIXED] Issue #180  : it is now possible to set the thermal limit with a dictionary


BDonnot (commented):
looks good

Issue #187, opened 2021-04-13
RedispReward returns reward greater than the reward_max


mlanden:
## Environment
 - Grid2op version: `1.4.0`
 - System: `Red Hat Enterprise Linux Server 7.9 (Maipo)"`
 - Additional system information

## Bug description
The redispReward is return rewards that are greater than the maximum reward calculated. 

## How to reproduce
Create environment with default reward and compare the rewards returned to the max reward computed in the initiative function. 
```

### Code snippet
<!--Expose the python code you want us to test-->
```python
import grid2op

env = grid2op.make()
_, reward, _, _ = env.step(env.action_space({}))

```
The output of the code snippet above
```
The reward returned is greater than the reward_max computed in initialize.

**Comment by BDonnot** [comment]
Hello,
Thanks for noticing, l will try to update it asap. 

It should not be a problem, maybe in the mean time you can have a look at other environments, that  should behave more normally i hope. 

The other env. close to the default one is
```python
import grid2op
env_name="l2rpn_case14_realistic"
env = grid2op.make(en_name) 
```

mlanden:
Unfortunately, that does not seem to be the case. with the ieee 14 bus, I still see
```
Max: 706.4000244140625, Reward: 1085.8973388671875 redispach 1085.8973388671875, overflow 0.0
```
where the overflow reward is what I'm adding, but it is not relevant here because it is 0. I am applying interpolation to this reward so the max should be accurate to get correct results from np.interp.

BDonnot:
For me, for later resuse:

```python
env_name="l2rpn_case14_sandbox"
env = grid2op.make(env_name)
obs = env.reset()
obs, reward, done, info = env.step(env.action_space())
print(f"obtained reward: {reward:.2f}")
print(f"max reward: {env.reward_range[1]:.2f}")
````

So indeed there is problem with the way the reward max is computed for this specific case.

Issue #185, opened 2021-04-11
Assertion Error: Trying to convert grid2op to GymEnv using grid2op.compat


vishrutmishra:
## Environment
 - Grid2op version: `1.5.0`
 - System: `windows 10`
 - Additional system information

## Bug description

Cannot convert grid2op to the gym environment using the code snippet given on https://grid2op.readthedocs.io/en/latest/gym.html

## How to reproduce

I tried to convert grip2op into a gym environment. But it gave us an assertion error.

### Command line

```bash
# no command line used.
```

### Code snippet

```python
import grid2op
from grid2op.gym_compat import GymEnv
env_name = "l2rpn_case14_sandbox"
env = grid2op.make(env_name)
gym_env = GymEnv(env)
```

## Current output

```
Traceback (most recent call last):
  File "D:\projects\cse574\demo\grid2op-demo.py", line 6, in <module>
    gym_env = GymEnv(env)
  File "C:\Python38\lib\site-packages\grid2op\gym_compat\gymenv.py", line 43, in __init__
    self.action_space = GymActionSpace(self.init_env)
  File "C:\Python38\lib\site-packages\grid2op\gym_compat\gym_act_space.py", line 131, in __init__
    self._fill_dict_act_space(dict_,
  File "C:\Python38\lib\site-packages\grid2op\gym_compat\gym_act_space.py", line 256, in _fill_dict_act_space
    my_type = SpaceType(low=low, high=high, shape=shape, dtype=dt)
  File "C:\Python38\lib\site-packages\gym\spaces\box.py", line 30, in __init__
    assert np.isscalar(low) and np.isscalar(high)
AssertionError
```

## Expected output

```
No errors
```

BDonnot:
Hello,

Thanks for the detailed issue.

Unfortunately, i could not reproduce it on my computer, neither locally (on a windows 10 nor on a linux machine) nor on colab using one of the notebook in the google colab website, see for example https://colab.research.google.com/github/rte-france/Grid2Op/blob/master/getting_started/00_SmallExample.ipynb

Can you run this example:

```python
import sys
# to give extra info on the python version
print(sys.version)
# for extra info on the platform
print(sys.platform)

import grid2op
# extra info on grid2op
print(grid2op.__version__)

# to make sure you have the latest version of the environment
grid2op.update_env()

# now create it
env = grid2op.make("l2rpn_case14_sandbox")
# "make sure" it works
obs = env.reset()

# and uses the gym_compat module
[...]

Issue #166, opened 2020-12-22
Does SingleEnvMultiProcess support simulate function?


hlhsu:
## Is your feature request related to a problem? Please describe.
I am trying to run the code under multi-process and I glad to know that we can handle it via SingleEnvMultiProcess. However, if I would like to call obs.simulate(action) like what we do in one process, it shows raise NoForecastAvailable("No forecasts are available for this instance of BaseObservation (no action_space "
grid2op.Exceptions.ObservationExceptions.NoForecastAvailable: Grid2OpException NoForecastAvailable "No forecasts are available for this instance of BaseObservation (no action_space and no simulated environment are set).

## Describe the solution you'd like
Could we also have the same implementation like multi_obs[process_id].simulate[action[process_id]?

## Additional context
Add any other context about the feature request here.

BDonnot:
Hello,

Unfortunately no it's not possible at the moment as stated in the documentation (admitedly a bit hidden):
https://grid2op.readthedocs.io/en/latest/environment.html?highlight=BaseMultiProcessEnvironment#grid2op.Environment.BaseMultiProcessEnvironment

> A broader support of regular grid2op environment capabilities as well as support for grid2op.Observation.BaseObservation.simulate() call might be added in the future.

It's not done yet because it's unclear how it would work.
Using multiEnv you receive a list of observations, what would it mean to do a "simulate" on this list ?
Do that mean you simulate the effect of the same action on all the observations of the list ? Or that you pass a list of actions ?

It's not very clear but if you have a use case that you can illustrate with a few lines of code, we will study how we could implement it :-)

hlhsu:
Thank you for your response. For example, we have 8 cores in a single machine. Then we can use SingleEnvMultiProcess to run 8 
different chronics at the same time (different observations). Then I expect that we can use the simulate function to simulate the effect of 8 actions corresponding to these 8 observations. Therefore, it just one action for one observation, and then we can train it with 8 observations at the same time. It's just for saving time.

bwitherspoon:
I would like to explore this enhancement also. I can currently work around it with an Actor/Learner like framework designed for multiple machines where I create N processes with N actors each with its own single process environment so I can simulate. This is a bit overkill for a single machine though because each actor is perfectly capable if generating batches of actions (and batches of actions it wants to simulate). I need to be able to generate as much experience as possible in parallel with the ability to simulate with forecasts.

[...]

PR #189, opened 2021-04-15
Ready for 1.5.1

BDonnot:
- [FIXED]: Issue #187 : improve the computation and the
  documentation of the `RedispReward`. This has an impact on the `env.reward_range` of all environments using this
  reward, because the old "reward_max" was not correct.
- [FIXED] Issue #181 : now environment can be created with
  a layout and a warning is issued in this case.
- [FIXED] Issue #180 : it is now possible to set the thermal
  limit with a dictionary
- [FIXED] a typo that would cause the attack to be discarded in the runner in some cases (cases for now not used)
- [FIXED] an issue linked to the transformation into gym box space for some environments,
  this **might** be linked to Issue #185
- [ADDED] a feature to retrieve the voltage angle (theta) in the backend (`backend.get_theta`) and in the observation.
- [ADDED] support for multimix in the GymEnv (lack of support spotted thanks to Issue #185)
- [ADDED] basic documentation of the environment available.
- [ADDED] Issue #166: support for simulate in multi environment
  settings.
- [IMPROVED] extra layer of security preventing modification of `observation_space` and `action_space` of environment
- [IMPROVED] better handling of dynamically generated classes
- [IMPROVED] the documentation of the opponent


BDonnot (commented):
looks good

Issue #196, opened 2021-04-20
Unable to run l2rpn_neurips_2020_track1_small using Rllib integration code 


rrhossain:
## Environment
 - Grid2op version: `1.5.1`
 - System: ubuntu 20.04 / MAC Os`
 - Additional system information

## Bug description
**The code is working for l2rpn-sandbox but it is producing error for l2rpn-neurips-track-1-small environment.**

### Code snippet
```
import gym
import ray
import gym
import numpy as np
class MyEnv(gym.Env):
    def __init__(self, env_config):
        import grid2op
        from grid2op.gym_compat import GymEnv
        from grid2op.gym_compat import ScalerAttrConverter, ContinuousToDiscreteConverter, MultiToTupleConverter
        # 1. create the grid2op environment
        if not "env_name" in env_config:
            raise RuntimeError("The configuration for RLLIB should provide the env name")
        nm_env = env_config["env_name"]
        del env_config["env_name"]
        self.env_glop = grid2op.make(nm_env, **env_config)

        # 2. create the gym environment
        self.env_gym = GymEnv(self.env_glop)
        obs_gym = self.env_gym.reset()

        # 3. (optional) customize it (see section above for more information)
        ## customize action space
        self.env_gym.action_space = self.env_gym.action_space.ignore_attr("set_bus").ignore_attr("set_line_status")
        self.env_gym.action_space = self.env_gym.action_space.reencode_space("redispatch",
                                                                             ContinuousToDiscreteConverter(nb_bins=11)
                                                                             )
        self.env_gym.action_space = self.env_gym.action_space.reencode_space("change_bus", MultiToTupleConverter())
        self.env_gym.action_space = self.env_gym.action_space.reencode_space("change_line_status",
                                                                             MultiToTupleConverter())
        self.env_gym.action_space = self.env_gym.action_space.reencode_space("redispatch", MultiToTupleConverter())
        ## customize observation space
        ob_space = self.env_gym.observation_space
        ob_space = ob_space.keep_only_attr(["rho", "gen_p", "load_p", "topo_vect", "actual_dispatch"])
        ob_space = ob_space.reencode_space("actual_dispatch",
                                           ScalerAttrConverter(substract=0.,
                                                               divide=self.env_glop.gen_pmax
                                                               )
                                           )
        ob_space = ob_space.reencode_space("gen_p",
                                           ScalerAttrConverter(substract=0.,
                                                               divide=self.env_glop.gen_pmax
                                                               )
                                           )
[...]

PR #198, opened 2021-04-26
Adressing issue 196 and 185

BDonnot:
- [FIXED]: #196 an issue related to the low / high of the observation if using the gym_compat module. Some more protections are enforced now.
- [FIXED]: #196 an issue related the scaling when negative numbers are used (in these cases low / max would be mixed up)
- [IMPROVED]: on windows at least, grid2op does not work with gym < 0.17.2 Checks are performed in order to make sure the installed open ai gym package meets this requirement (see issue #185 )
- [IMPROVED] the seed of openAI gym for composed action space (see issue `https://github.com/openai/gym/issues/2166`): waiting for an official fix, grid2op will use the solution proposed there https://github.com/openai/gym/issues/2166#issuecomment-803984619 )


BDonnot (commented):
if tests pass this looks good

PR #200, opened 2021-05-03
Add few new features, fix some others

BDonnot:
[BREAKING]: allow the opponent to chose the duration of its attack. This breaks the previous "Opponent.attack(...)"
    signature by adding an object in the return value. All code provided with grid2op are compatible with this
    new change. (for previously coded opponent, the only thing you have to do to make it compliant with
    the new interface is, in the opponent.attack(...) function return whatever_you_returned_before, None instead
    of simply whatever_you_returned_before
    [FIXED]: an issue with the IncreasingFlatReward reward types
    [ADDED]: a reward EpisodeDurationReward that is always 0 unless at the end of an episode where it returns a float
    proportional to the number of step made from the beginning of the environment.
    [ADDED]: in the Observation the possibility to retrieve the current number of steps


BDonnot (commented):
Looks good

PR #227, opened 2021-06-13
Notebook 02,03,05,06 edits

JochenC:
several minor language edits: 

Also, these following still need to change, and being discussed: 

Notebook 02:
Some internal notes could be replaces, like:
-	“Hopefully, some utilities functions in grid2op allows to build these graphs rapidly.”

Notebook 03:
-	Not sure if ambiguity is the right word to choose. Ambiguity means if something is not unique, while I think a better term to use would be “infeasible” or “not valid”, from the description, this is what I understand you aim to say.
-	Not sure what this sentence means “Sometimes it can also be interesting to reason at the substation level, and to explicitly say we want to modify this substation and not anything else.”. What does “reason” mean in this context?

Notebook 05:
-	Please change the first 3 links, they are broken, and the last line link is broken, too

Notebook 06:
-	“"automatons" that automatically transform an proposed redispatching action into a valid redispatching action. These automatons”, it may not be clear for everyone what an automaton is.
-	“Any attempt to dispatch a generator that is not dispatchable leads to an ambiguous action.” Not clear how that works. I think the word “ambigious” is here not correctly used. It would be better to say “not valid” or “infeasible” unless I am missing the point. I would propose to check this point. The same is a few more times along this notebook.


BDonnot:
Hello,

Thanks a lot for all these modifications. To answer some of your comments:

> Notebook 02:
Some internal notes could be replaces, like:
    “Hopefully, some utilities functions in grid2op allows to build these graphs rapidly.”

Then please go ahead, change what you think is best :-)

> Not sure if ambiguity is the right word to choose. Ambiguity means if something is not unique, while I think a better term to use would be “infeasible” or “not valid”, from the description, this is what I understand you aim to say.

I think i see what you mean, but unfortunately "ambiguous" is used a lot in grid2op, in the notebook, in the doc, in the code. Changing everything would take too much time.

Originally, it was introduced to deal with case like "i want to set this element to bus 2 and change the bus of this element" which is a behaviour that is ambiguous (what does the agent meant there ?) but i agree now some other concepts have been added, so the word "ambiguous" might not be that suited.

> Sometimes it can also be interesting to reason at the substation level, and to explicitly say we want to modify this substation and not anything else

Sorry, bad translation from a French sentence, the word is not used properly. What i meant by "reason" in this context was "think, "look explicitely". I hope it's more clear.

> Please change the first 3 links, they are broken, and the last line link is broken, too

Sorry i won't have time before the beginning of the competition. So maybe we need to keep them as is if it does not make the things too hard to read. Maybe they can also be fixed easily, i don't remember which links are there.

> “"automatons" that automatically transform an proposed redispatching action into a valid redispatching action. These automatons”, it may not be clear for everyone what an automaton is.

You are totally correct, maybe we can try to clarify this in one or two sentence (for this specific automaton, it's similar to first and secondary frequency control automatons, but I did not felt like explaining everything there...)

> “Any attempt to dispatch a generator that is not dispatchable leads to an ambiguous action.” Not clear how that works. I think the word “ambigious” is here not correctly used. It would be better to say “not valid” or “infeasible” unless I am missing the point. I would propose to check this point. The same is a few more times along this notebook.

Again you are correct: "ambiguous" here is not the english word, but the grid2op concept: an action is ambiguous if the environment does not know how to process it.

An action is invalid if the rules prevents to do it (which is a different concept). 

For example, changing the status of 2 powerlines can be invalid depending on the `parameters` of the environment. "Set an object to bus 2 and change its bus" is ambiguous, regardless of anything.

This concept of "ambiguous" has then been "extended" to cases where the action was "always ambiguous / illformed / misspecified", but i totally agree that this is not clear at all and  confusing. As previously stated, changing "ambiguous" now for a better alternative would meaning changing lots of code, documentation and notebook, and i honestly don't have the time :-( But for sure it would be way better !

JochenC:
@BDonnot Thank you very much for your comments. I made the above changes and committed them here.

BDonnot:
I'm waiting the results of the tests. If they pass i will merge these changes.

Issue #245, opened 2021-08-20
Powerline status are not properly taken into account in "obs.simulate" right after a powerline is disconnected by protection


BDonnot:
## Environment
 - Grid2op version: `1.6.2.post1`
 - System: `ubuntu 20.04`
 - Additional system information

## Bug description

see title
## How to reproduce

See the example bellow

### Command line

```bash
# command line used if any 
```
 Just run the above code in a python repl and the bug will appear.

### Code snippet

The code snippet (see the test_issue_245.py) is:

```python
import grid2op
import warnings

import grid2op
from grid2op.tests.helper_path_test import *
from grid2op.Action import CompleteAction

class Issue245Tester(unittest.TestCase):
    def setUp(self) -> None:
        with warnings.catch_warnings():
            warnings.filterwarnings("ignore")
            env_nm = "l2rpn_case14_sandbox"
            self.env = grid2op.make(env_nm,
                                    test=True,
                                    action_class=CompleteAction
                                    )
            self.env.seed(0)
            self.env.reset()

    def test_simulate(self):
        self.env.set_thermal_limit(2.0 * np.array([162.49806 , 127.40141 , 107.29365 , 110. / 2., 107.90302 ,
                                                   48.58299 , 135.80658 , 434.52097 , 252.74222 , 649.2642  ,
                                                   56.013855, 183.57967 , 324.4969  ,  79.42569 , 319.6616  ,
                                                   100.69238 ,  46.787445,  73.96375 , 895.7249  , 563.3386  ],
                                                  dtype=dt_float)
                                   )
        obs = self.env.reset()
        assert obs.line_status[3]
        obs, *_ = self.env.step(self.env.action_space())
        assert obs.line_status[3]
        obs, *_ = self.env.step(self.env.action_space())
        assert obs.line_status[3]
        obs, *_ = self.env.step(self.env.action_space())
        assert not obs.line_status[3]
        # line 3 is disconnected, now i try to simulate an action that reconnects it (it should not work) !
        act = self.env.action_space({"set_bus": {"lines_or_id": [(3, 1)]}})
        *_, sim_info = obs.simulate(act)
        assert len(sim_info["exception"]) > 0, "there should be an exception because the action is illegal"
        *_, info = self.env.step(act)
        assert len(info["exception"]) > 0, "there should be an exception because the action is illegal"
```

## Current output

The current output is:

```
======================================================================
FAIL: test_simulate (test_issue_245.Issue245Tester)
[...]

PR #246, opened 2021-08-21
upgrade to version 1.6.3

BDonnot:
- [FIXED] a bug that allowed to use wrongly the function `backend.get_action_to_set()` even when the backend
  has diverged (which should not be possible)
- [FIXED] a bug leading to non correct consideration of the status of powerlines right after the activation
  of some protections (see `Issue#245 <https://github.com/rte-france/Grid2Op/issues/245>`_ )
- [IMPROVED] the PandaPowerBackend is now able to load a grid with a distributed slack bus. When loaded though, the
  said grid will be converted to one with a single slack bus (the first slack among the distributed)
- [IMPROVED] massive speed-ups when copying environment or using `obs.simulate` (sometimes higher than 30x speed up)
- [IMPROVED] **experimental** compatibility with different frameworks thanks to the possibility to serialize, as text
  files the class created "on the fly" (should solve most of the "pickle" error). See `env.generate_classes()`
  for an example usage. Every feedback is appreciated.


BDonnot (commented):
Looks good if tests pass

BDonnot:
Wrong branch...

Issue #255, opened 2021-10-07
Dangling references prevent GC to collect instances


vakker:
## Environment
 - Grid2op version: `1.6.3`
 - System: `ubuntu18.04`

## Bug description
Then the env instance is deleted (e.g. by `del`) then GC should collect it. 
It does happen in most scenarios, but in some cases there are references to the env instance that doesn't allow for the collection of obsolete instances. 

## How to reproduce
Running the following shows that GC does it's job:
```python
import gc
import sys

import grid2op


class TestClass:
    pass


def main():
    for i in range(50):
        env = grid2op.make('l2rpn_icaps_2021_small')
        d = TestClass()
        print(
            'Grid2Op instances:',
            len([
                o for o in gc.get_objects()
                if isinstance(o, grid2op.Environment.BaseEnv)
            ]))
        print('Test instances:',
              len([o for o in gc.get_objects() if isinstance(o, TestClass)]))
        # print('Ref', sys.getrefcount(d))
        # print('Refs', gc.get_referrers(d))
        del d
        del env
        gc.collect()
        print(
            'Grid2Op instances:',
            len([
                o for o in gc.get_objects()
                if isinstance(o, grid2op.Environment.BaseEnv)
            ]))
        print('Test instances:',
              len([o for o in gc.get_objects() if isinstance(o, TestClass)]))
        print('#########')


main()
```
However, if there's a wrapper around the env (as it's required for RLlib) there are calls to `env.step` and the wrapper (or other parts of the framework) might store a reference to the env through the returned observation, reward or info. 

In my case using `deepcopy` (see below) resolves the issue:
```python
obs_grid, reward, done, info = copy.deepcopy(self.env_grid.step(action_grid))
```

Otherwise, there are a pile of dangling instances that doesn't get collected. 
Currently I'm using the following (that's called in the `reset` method of my wrapper when it's needed e.g. for curriculum learning):
```python
    def _reset_env(self):
        self.env_grid.close()

        self.env_grid = None
        self._init_all()  # runs grid2op.make 
        self.episodes = 0

        gc.collect()

        # print(
        #     'Grid2Op instances:',
        #     len([
        #         o for o in gc.get_objects()
        #         if isinstance(o, grid2op.Environment.BaseEnv)
        #     ]))
```

The commented out bit prints the number of grid2op env instances in the given process.

BDonnot:
Hello, 
[...]

PR #257, opened 2021-11-05
Upgrade to version 1.6.4

BDonnot:
- [BREAKING] the name of the python file for the "agent" module are now lowercase (complient with PEP). If you
  did things like `from grid2op.Agent.BaseAgent import BaseAgent` you need to change it like
  `from grid2op.Agent.baseAgent import BaseAgent` or even better, and this is the preferred way to include
  them: `from grid2op.Agent import BaseAgent` It should not affect lots of code.
- [FIXED] a bug where the shunt had a voltage when disconnected using pandapower backend
- [FIXED] a bug preventing to print the action space if some "part" of it had no size (empty action space)
- [FIXED] a bug preventing to copy an action properly (especially for the alarm)
- [FIXED] a bug that did not "close" the backend of the observation space when the environment was `closed`. This 
  might be related to #255
- [ADDED] serialization of `current_iter` and `max_iter` in the observation.
- [ADDED] the possibility to use the runner only on certain episode id
  (see `runner.run(..., episode_id=[xxx, yyy, ...])`)
- [ADDED] a function that returns if an action has any change to modify the grid see `act.can_affect_something()`
- [ADDED] a ttype of agent that performs predefined actions from a given list
- [ADDED] basic support for logging in environment and runner (more coming soon)
- [ADDED] possibility to make an environment with an implementation of a reward, instead of relying on a reward class.
- [ADDED] a possible implementation of a N-1 reward
- [IMPROVED] right time stamp is now set in the observation after the game over.
- [IMPROVED] correct current number of steps when the observation is set to a game over state.
- [IMPROVED] documentation to clearly state that the action_class should not be modified.
- [IMPROVED] possibility to tell which chronics to use with the result of `env.chronics_handler.get_id()` (this is also
  compatible in the runner)
- [IMPROVED] it is no more possible to call "env.reset()" or "env.step()" after an environment has been closed: a clean error
  is raised in this case.


BDonnot (commented):
Looks good if tests pass

PR #299, opened 2022-04-02
fix two small malformations

mowang111:


BDonnot (approved):
Thanks for the fix :-)

Issue #327, opened 2022-06-01
Error creating new variables using as_networkx()


EloyAnguiano:
## Environment
 - Grid2op version: `1.7.0`        
 - Stable Baselines 3 version: `1.4.0`        
 - System: `Ubuntu 18.04 LTS`
 - Python `3.9.12`

## Bug description
I'm trying to add some features to the space using the as_networkx method but sometimes there are some problems with that conversion.

### Command line
```bash
# python script.py
```

### Code snippet
```python
import torch as th
from stable_baselines3 import SAC
from stable_baselines3.common.torch_layers import BaseFeaturesExtractor
import gym
from grid2op import make
from grid2op.gym_compat import BoxGymActSpace, GymEnv
from copy import deepcopy
import networkx as nx
from gym.spaces import Box
import numpy as np
import torch.nn as nn

def build_matrix(obs, key):
    if isinstance(obs, nx.Graph):
        graph = obs
    else:
        graph = obs.as_networkx()

    matrix = nx.attr_matrix(graph, edge_attr=key)[0]
    return matrix

class FeatureExtractor(BaseFeaturesExtractor):
    def __init__(self, observation_space:  gym.spaces.Dict, features_dim: int, device):
        super(FeatureExtractor, self).__init__(
            observation_space, features_dim)

        self.device = device

        space = observation_space["rho_matrix"]
        n = space.shape[0] * space.shape[1]

        self.last_dense = nn.Sequential(
            nn.Linear(n, self.features_dim),
            nn.ReLU())


    def forward(self, observations: gym.spaces.Dict) -> th.Tensor:
        return self.last_dense.forward(observations["rho_matrix"].flatten(1))

POLICY = {  "features_extractor_class": FeatureExtractor,
            "features_extractor_kwargs": {
                "features_dim": 256,
                "device": "cpu",
            },
            "activation_fn": th.nn.ReLU}
if __name__ == "__main__":

    grid2op_env = make("rte_case5_example", test=True)

    env = GymEnv(grid2op_env)  # create the gym environment

    # CHANGING ACTION SPACE
    env.action_space = BoxGymActSpace(env.init_env.action_space)

    # CHANGING OBSERVATION SPACE
    # Adding vars
    grid2op_obs = grid2op_env.reset()

    # Adding vars
    graph = deepcopy(grid2op_obs).as_networkx()

    # graph_vars = {}
    graph_vars = {"rho":(-1000, 1000)}

    for k in graph_vars.keys():
[...]

Issue #321, opened 2022-05-20
obs.curtailment_limit_effective documentation is unclear


EBoguslawski:
## Documentation issue description

If you run the following code which simply executes a curtailement action and print two quantities :

```
import grid2op
from grid2op.Parameters import Parameters
from grid2op.Action import PlayableAction

# Create environement
p=Parameters()
p.LIMIT_INFEASIBLE_CURTAILMENT_STORAGE_ACTION = True
env_name  = "educ_case14_storage"
env = grid2op.make(env_name, test=True, action_class=PlayableAction, param=p)

# Initialize the environment
env.seed(0)
env.set_id(0)
env.reset()

# Apply a curtailment action
action = env.action_space({'curtail': [(2, 0.1), (3, 0.15), (4, 0.2)]})
obs, _, _, _ = env.step(action)

# Print two quantites
print("obs.curtailment_limit :", obs.curtailment_limit)
print("obs.curtailment_limit_effective :", obs.curtailment_limit_effective)
```
You obtain the following output : 

```
obs.curtailment_limit : [1.   1.   0.1  0.15 0.2  1.  ]
obs.curtailment_limit_effective : [1.  1.  0.1 0.  0.  1. ]
```
As you can see effective curtailment limits seem to be more strict than the initial curtailment limits for gen 3 and 4. This is strange because the LIMIT_INFEASIBLE_CURTAILMENT_STORAGE_ACTION parameter should on the contrary soften the effective curtailment action.

This doesn't match the actual documention for `curtailment_limit_effective`  which is "Limit (in ratio of gen_pmax) imposed on each renewable generator effectively imposed by the environment."

## Suggested modifications

Precise in the documentation how `curtailment_limit_effective` is calculated.

BDonnot:
Thanks 😊

Can you check the code and propose a good wording if it's possible?

Thanks

BDonnot:
(this has been addressed and will be part of grid2op 1.7.1 release)

BDonnot:
(I re open it as it is not yet merged into the main branch, i'll close it when i'll release grid2op version 1.7.1)

PR #319, opened 2022-05-16
Line Disconnection bug in _BackendAction fix

anton-musrevinu:
Due to an indexing error int the update_state method of the _BackendAction lines with idx 0 and -1(and sometimes 1) are sometimes disconnected.

A script to reproduce the bug is given below. Here a seed was found that produces the bug after a few steps.
[reproduce_stepping_bug.zip](https://github.com/rte-france/Grid2Op/files/8698834/reproduce_stepping_bug.zip)

Issue #281, opened 2022-01-25
Missing some keys when using GymEnv


BDonnot:
## Environment
 - Grid2op version: `1.6.5`
 - System: `all`
 - Additional system information

## Bug description

When setting the action class to be `CompleteAction` we cannot use the `GymEnv` because of some missing keys.

## How to reproduce

Just run this in a python repl:

```python
import grid2op
from grid2op.Action import CompleteAction
from grid2op.gym_compat import GymEnv

if __name__ == "__main__":
    env = grid2op.make("educ_case14_storage",
                        test=True,
                        action_class=CompleteAction)

    env_gym = GymEnv(env)
```

## Current output

```
Traceback (most recent call last):
  File "test_box_act copy.py", line 15, in <module>
    env_gym = GymEnv(env)
  File "grid2op/gym_compat/gymenv.py", line 45, in __init__
    self.action_space = GymActionSpace(self.init_env)
  File "/grid2op/gym_compat/gym_act_space.py", line 138, in __init__
    dict_ = self._fix_dict_keys(dict_)
  File "/grid2op/gym_compat/gym_act_space.py", line 273, in _fix_dict_keys
    res[self.keys_grid2op_2_human[k]] = v
KeyError: 'shunt_p'
```

## Expected output

```
No error expected
```

PR #329, opened 2022-06-03
Upgrade to version 1.7.1

BDonnot:
Breaking changes
---------------------------
- The possibility to propagate keyword arguments between the environment
  and the runner implied adding some arguments in the constructor of 
  `PandapowerBackend`. So if you made a class that inherit from it, you should
  add these arguments in the constructor (otherwise you will not be able to use
  the runner) [This should not impact lot of codes, if any]

Fixed issues
------------------
- a documentation issue https://github.com/rte-france/Grid2Op/issues/281
- a bug preventing to use the `FromChronix2grid` chronics class when 
  there is an opponent on the grid.
- a documentation issue https://github.com/rte-france/Grid2Op/issues/319
  on notebook 11
- some issues when the backend does not support shunts data (caused during the
  computation of the size of the observation) Tests are now performed in
  `grid2op/tests/test_educpp_backend.py`
- a bug when downloading an environment when the archive name is not the 
  same as the environment names (attempt to delete a non existing folder). This 
  is the case for `l2rpn_wcci_2022` env. For this env, your are forced to use
  grid2op version >= 1.7.1
- an issue when converting a "done" action as a graph, see
  https://github.com/rte-france/Grid2Op/issues/327

New features
---------------------------
- score function for the L2RPN WCCI 2022 competition

Improvments
---------------------------

- adding the compatibility with logger in the reward functions.
- when there is a game over caused by redispatching, the observation is
  not updated, as it is the case for other type of game over (improved consistency)
- it is now possible to make an environment with a backend that
  cannot be copied.
- the arguments used to create a backend can be (if used properly)
  re used (without copy !) when making a `Runner` from an environment for example.
- description and definition of `obs.curtailment_limit_effective` are now
  consistent (issue https://github.com/rte-france/Grid2Op/issues/321)


BDonnot (commented):
looks good

PR #335, opened 2022-06-16
Creation of subgrid objects finished

Farid-Najar:


grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -591,22 +591,30 @@ def _verify_domains(self, domains : MADict) :
...
-                    raise DomainException(f"The id must be of type int. Type {type(sub_id)} is not valid")
+                    raise DomainException(f"{agent} : The id must be of type int. Type {type(sub_id)} is not valid")
                 if sub_id < 0 or sub_id > len(self._cent_env.name_sub) :
-                    raise DomainException(f"The substation's id must be between 0 and {len(self._cent_env.name_sub)-1}, but {sub_id} has been given")
-    
+                    raise DomainException(f"{agent} : The substation's id must be between 0 and {len(self._cent_env.name_sub)-1}, but {sub_id} has been given")
+            sum_subs += len(domains[agent])
+
+        if sum_subs != self._cent_env.n_sub:
+            raise DomainException(f"The sum of sub id lists' length must be equal to _cent_env.n_sub = {self._cent_env.n_sub} but is {sum_subs}")
```
BDonnot, line 617: `f"agent id {agent} :`

would be better to start the error message

grid2op/multi_agent/test_ma.py
```diff
@@ -49,12 +49,35 @@ def test_verify_domains(self):
         observation_domains = action_domains
         try:
             MultiAgentEnv(self.env, observation_domains, action_domains)
+            assert False
+            
+        except DomainException :
+            assert True
```
BDonnot, line 55: Whaaaaat is that ? No no no, raise proper error and have a proper handling mechanism. What's the point to assert True or False ?

BDonnot: You can use [self.assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) for example

grid2op/multi_agent/test_ma.py
```diff
@@ -49,12 +49,35 @@ def test_verify_domains(self):
...
+            'agent_0' : [0],
+            'agent_1' : [5,6,7,8,9,10,11,12,13]
+        }
+        observation_domains = action_domains
+        try:
+            MultiAgentEnv(self.env, observation_domains, action_domains)
+            assert False
             
         except DomainException :
             assert True
```
BDonnot, line 67: Same here, do not assert True or assert False

grid2op/multi_agent/test_ma.py
```diff
@@ -49,12 +49,35 @@ def test_verify_domains(self):
...
+            'agent_0' : [],
+            'agent_1' : list(range(self.env.n_sub))
+        }
+        observation_domains = action_domains
+        try:
+            MultiAgentEnv(self.env, observation_domains, action_domains)
+            assert False
+            
+        except DomainException :
+            assert True
```
BDonnot, line 79: don't assert True or False

grid2op/multi_agent/test_ma.py
```diff
@@ -148,7 +171,44 @@ def test_build_subgrid_obj(self):
...
+            'agent_0' : self.action_domains['agent_1'],
+            'agent_1' : self.action_domains['agent_0']
+        }
+        # run redispatch agent on one scenario for 100 timesteps
+        ma_env = MultiAgentEnv(self.env, observation_domains, action_domains)
+        
+        assert ma_env._subgrids_cls['action']['agent_0'].n_gen +  ma_env._subgrids_cls['action']['agent_1'].n_gen == self.env.n_gen
+        assert ma_env._subgrids_cls['action']['agent_0'].n_load +  ma_env._subgrids_cls['action']['agent_1'].n_load == self.env.n_load
+        assert ma_env._subgrids_cls['action']['agent_0'].n_interco ==  ma_env._subgrids_cls['action']['agent_1'].n_interco
+        assert ma_env._subgrids_cls['action']['agent_0'].n_line \
```
BDonnot, line 190: It's maybe made in a further test, but you need to test that these attributes are class attributes and not just member attributes

grid2op/multi_agent/test_ma.py
```diff
@@ -148,7 +171,44 @@ def test_build_subgrid_obj(self):
...
+            + ma_env._subgrids_cls['action']['agent_1'].n_line \
+            + ma_env._subgrids_cls['action']['agent_0'].n_interco == self.env.n_line
```
BDonnot, line 211: need to check for line_or and line_ex also !

And for the shunts too

Also, make a function to test that instead of copy pasting it from 187-191

BDonnot:
Tons of tests missing, only about 10-20% of what I said is performed. Waiting for the remaining 80-90%

BDonnot (changes requested):
Tons of tests missing, only about 10-20% of what I said is performed. Waiting for the remaining 80-90%

grid2op/Action/BaseAction.py
```diff
@@ -14,6 +14,7 @@
 from grid2op.dtypes import dt_int, dt_bool, dt_float
 from grid2op.Exceptions import *
 from grid2op.Space import GridObjects
+from grid2op.Observation.baseObservation import BaseObservation
```
BDonnot, line 18: Why is that ? I'm not sure it's mandatory ?

grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -0,0 +1,1070 @@
...
+
+        Args:
+            observation (BaseObservation): _description_
+        """
+        if self.global_obs:
+            for agent in self.agents:
+                self.observations[agent] = observation.copy()
+        else:
+            #TODO
+            raise NotImplementedError("Local observations are not available yet !")
```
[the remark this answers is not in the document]
BDonnot: Also, here, you suppose that you need the observation of the full grid to build the sub observation.

Do you really ?

You can do this with only the state of the backend ? Why do you chose to do that this way ? What does it entails ? What are the pros and cons ?

grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -0,0 +1,1070 @@
...
+            
+            if len(domains[agent]) == 0:
+                raise DomainException(f"Agent id {agent} : The domain is empty !")
+            
+            for sub_id in domains[agent] : 
+                if not (isinstance(sub_id, int) 
+                        or isinstance(sub_id, dt_int) 
+                        or isinstance(sub_id, np.int32) 
+                        or isinstance(sub_id, np.int64)
+                ):
```
BDonnot, line 674: not a fan of this to be honest

grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -0,0 +1,1070 @@
...
+                    raise DomainException(f"Agent id {agent} : The id must be of type int. Type {type(sub_id)} is not valid")
+                if sub_id < 0 or sub_id > len(self._cent_env.name_sub) :
+                    raise DomainException(f"Agent id {agent} : The substation's id must be between 0 and {len(self._cent_env.name_sub)-1}, but {sub_id} has been given")
+            sum_subs += len(domains[agent])
+
+        if sum_subs != self._cent_env.n_sub:
+            raise DomainException(f"The sum of sub id lists' length must be equal to _cent_env.n_sub = {self._cent_env.n_sub} but is {sum_subs}")
+    
+    
+    def observation_space(self, agent):
```
BDonnot, line 684: be specific here:

does it expects and agent id (0, 1, etc.) or an agent name ?

grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -0,0 +1,1070 @@
...
+        """
+        Takes in agent and returns the observation space for that agent.
+
+        MUST return the same value for the same agent name
+
+        Default implementation is to return the observation_spaces dict
+        """
+        return self.observation_spaces[agent]
+
+    def action_space(self, agent):
```
BDonnot, line 694: same as above, does it expects an id or a name ?

grid2op/multi_agent/multi_agentExceptions.py
```diff
@@ -0,0 +1,16 @@
...
+# This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
+
+from grid2op.Exceptions.Grid2OpException import Grid2OpException
+
+class DomainException(Grid2OpException):
+    """This class represents the exceptions that occur in the
+    _verify_domains method of MultiAgentEnv. It occurs when 
+    the given domain by the user is not valid.
+    """
+    pass
```
BDonnot, line 16: add a new blank line at end of file

grid2op/multi_agent/subGridObjects.py
```diff
@@ -0,0 +1,50 @@
...
+        res["sub_orig_ids"] = cls.sub_orig_ids
+        res["local2global"] = cls.local2global
+        res["mask_sub"] = cls.mask_sub
+        res["mask_load"] = cls.mask_load
+        res["mask_gen"] = cls.mask_gen
+        res["mask_storage"] = cls.mask_storage
+        res["mask_line_or"] = cls.mask_line_or
+        res["mask_line_ex"] = cls.mask_line_ex
+        res["mask_shunt"] = cls.mask_shunt
+        res["agent_name"] = cls.agent_name
```
BDonnot, line 50: look at the grid2op code (GridObjects 3044) and do the same !

grid2op/multi_agent/utils.py
```diff
@@ -0,0 +1,151 @@
+# Copyright (c) 2019-2020, RTE (https://www.rte-france.com)
+# See AUTHORS.txt
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
+
+from dataclasses import dataclass
```
BDonnot, line 9: what is all this file doing ?

BDonnot (changes requested):
I will have a look at the tests later on, but there are some issues with the code as is (tests or not).


Make sure to fix them

grid2op/Space/GridObjects.py
```diff
@@ -2802,15 +2805,19 @@ def get_obj_substations(cls, _sentinel=None, substation_id=None):
...
-            "lines_or_id"
-        ]
+        try:
+            res[cls.line_or_to_sub_pos[dict_["lines_or_id"]], cls.LOR_COL] = dict_[
+                "lines_or_id"
+            ]
+        except Exception as exc_:
+            import pdb
+            pdb.set_trace()
+            print("toto")
```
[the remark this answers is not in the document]
Farid-Najar: I didn't add this part. It must come from the fetch.

grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -0,0 +1,1070 @@
...
+        #        gridobj = self._subgrids_cls['observation'][agent],
+        #        env = self._cent_env,
+        #        rewardClass=self._cent_env._rewardClass,
+        #        observationClass=self._cent_env._observationClass,
+        #        actionClass=self._cent_env._actionClass,
+        #        #TODO following parameters
+        #        with_forecast=True,
+        #        kwargs_observation=None,
+        #    )
+
```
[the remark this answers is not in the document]
Farid-Najar: Some comments could be interesting for future works as an idea to begin with. I'll remove others.

grid2op/multi_agent/utils.py
Farid-Najar: It gives the possibility to define an agent selector that gives priority to agents.

grid2op/multi_agent/subGridObjects.py
Farid-Najar: You mean 3348 ? The same class method ?

grid2op/multi_agent/multiAgentEnv.py
Farid-Najar: Agents are identified by their names/id. It's the same and it's a string. They are not identified by an integer.
I'll add ": str" to clarify that.

grid2op/multi_agent/multiAgentEnv.py
Farid-Najar: what do you suggest instead?

grid2op/multi_agent/multiAgentEnv.py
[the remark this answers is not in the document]
Farid-Najar: For global observations, I think it's faster to give a copy of the observation computed and given by the _cent_env.step. Otherwise, we'll do the exact same computation n_agents times!

grid2op/multi_agent/multiAgentEnv.py
BDonnot: what if someone uses an int type not listed there ? Do you pretend to know all python framework and all their int type ?

grid2op/multi_agent/subGridObjects.py
BDonnot: If I said 3044 then I probably meant 3044. Have you checked there what's going on ? I suppose not... You should....

grid2op/multi_agent/utils.py
BDonnot: But is it necessary right now ?

grid2op/Space/GridObjects.py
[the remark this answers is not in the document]
BDonnot: But you can remove it, right ? ...

grid2op/multi_agent/subGridObjects.py
Farid-Najar: I did check indeed, but I don't see the link with this part.

grid2op/multi_agent/utils.py
Farid-Najar: Yes. It gives the order of agents even in the simple case (c.f. __init__ and step for MultiAgentEnv).

PR #350, opened 2022-07-01
_build_global_action and tests

Farid-Najar:


grid2op/multi_agent/multiAgentEnv.py
```diff
@@ -152,6 +152,16 @@ def __init__(self,
...
+        self.done = dict(
+            zip(
+                self.agents, [False for _ in range(self.num_agents)]
+            )
+        )
+        self.info = dict(
+            zip(
+                self.agents, [{} for _ in range(self.num_agents)]
+            )
+        )
```
BDonnot, line 164: nothing to do here...

Do you at least pay attention to what we said and what you are doing ?

grid2op/tests/test_ma.py
```diff
@@ -22,15 +23,22 @@
...
-
         self.action_domains = {
             'agent_0' : [0,1,2,3, 4],
             'agent_1' : [5,6,7,8,9,10,11,12,13]
         }
+        
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            
+            custom_params = Parameters()
```
BDonnot, line 34: nope... 

custom_params = self.cent_env.parameters

Why do you start from scratch ?

grid2op/tests/test_ma.py
```diff
@@ -906,6 +917,76 @@ def test_local_action_to_global_set_storage(self):
...
+        # do nothing
+        action = dict()
+        local_id = 0
+        for agent in self.ma_env.agents:
+            action[agent] = self.ma_env.action_spaces[agent]({})
+            action[agent].change_line_status = [local_id, local_id+1]
+        
+        self.ma_env._build_global_action(action, self.ma_env.agents)
+        
+        ref_action = self.ma_env._cent_env.action_space({})
```
BDonnot, line 953: would be easier to read if it was called "do nothing"...

grid2op/multi_agent/multiAgentEnv.py
Farid-Najar: Done is not used but info is used in illegal action handling

grid2op/tests/test_ma.py
Farid-Najar: done

PR #352, opened 2022-07-01
adding base things for Farid to work on multi agent

BDonnot:
add an example for multi random agent, fix the print issue


BDonnot (commented):
looks good, wait for test on bdonnot repo

Issue #393, opened 2023-01-09
Error when creating a new environment with LightSimBackend


AvisP:
## Environment
 - Grid2op version: `1.8.0`
 - System: `macos 12.6.2 `
 - Python: `3.10.7`

## Bug description
When trying to create a gym environment and using LightSimGrid as backend, there is an error produced but when using Pandapower as backend there is no error.

## How to reproduce
Running the following set of code will produce the error
```
import os
import sys
from lightsim2grid.lightSimBackend import LightSimBackend
from grid2op.gym_compat import GymEnv
import grid2op
from gym import Env
from gym.utils.env_checker import check_env
try:
#     from lightsim2grid import LightSimBackend
    bk_cls = LightSimBackend
except ImportError as exc:
    print(f"Error: {exc} when importing faster LightSimBackend")
# from grid2op.Backend import PandaPowerBackend
# bk_cls_panda = PandaPowerBackend
    
env_name = "l2rpn_case14_sandbox"
training_env = grid2op.make(env_name, backend=bk_cls())  # we put "test=True" in this notebook because...
# it's a notebook to explain things. Of course, do not put "test=True" if you really want
# to train an agent...
gym_env = GymEnv(training_env)
```

The error message generated is
```
Traceback (most recent call last):
  File "/Users/paula/Desktop/Projects/RL Practice/L2RPN Aspen/Demo Notebooks/LightSimBackend_test.py", line 22, in <module>
    training_env = grid2op.make(env_name, backend=bk_cls())  # we put "test=True" in this notebook because...
  File "/Users/paula/Desktop/Projects/venvs/L2PRN_G2O_1_8_0/lib/python3.10/site-packages/grid2op/MakeEnv/Make.py", line 427, in make
    return make_from_path_fn(
  File "/............./L2PRN_G2O_1_8_0/lib/python3.10/site-packages/grid2op/MakeEnv/MakeFromPath.py", line 764, in make_from_dataset_path
    env = Environment(
  File "/............./venvs/L2PRN_G2O_1_8_0/lib/python3.10/site-packages/grid2op/Environment/Environment.py", line 161, in __init__
    self._init_backend(
  File "/............../venvs/L2PRN_G2O_1_8_0/lib/python3.10/site-packages/grid2op/Environment/Environment.py", line 396, in _init_backend
    *_, fail_to_start, info = self.step(do_nothing)
[...]

PR #395, opened 2023-01-11
Update to version 1.8.1

BDonnot:
Fixed issues
------------------
- a deprecation with numpy>= 1.24 (**eg** `np.bool` and `np.str`), related to #393 

Additions
--------------
- the baseAgent class now has two new template methods `save_state` and `load_state` to save and
  load the agent's state during Grid2op simulations. Examples can be found in L2RPN baselines (PandapowerOPFAgent and curriculumagent).

Improvments
-------------------
- improved error messages in pandapower backend when the grid do not converge due to disconnected generators or loads.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,13 +31,16 @@ Change Log
 - [???] "asynch" multienv
 - [???] properly model interconnecting powerlines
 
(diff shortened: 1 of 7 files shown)


BDonnot (commented):
Looks good if tests pass

PR #398, opened 2023-01-17
support gym rgb_array rendering

wduguay-air:
### Proposed changes
- Support gym's `rgb_array` render mode. This returns a numpy array of rgb values for all pixels.
- The `mode` argument of `GymEnv.render()` is soon to be deprecated by gym. It causes a warning at every render. This is replaced by a `render_mode` attribute.


BDonnot:
Hello, 

For information, you can use it if you install the "dev" version of grid2op with :
`pip install Grid2Op==1.8.2.dev0`

PR #457, opened 2023-05-25
dev_1.8.2  : add default rules to be set by area

GoubetClem:
Add DefaultRules to be set for a list of list of substations ids. 
gamerules_class RulesByArea: when called, roll down a list of default rules. Use a custom LookParam function to check number of lines and substations allowed to have their status changed.
Basically look for MAX_LINE_STATUS_CHANGED and MAX_SUB_CHANGED and allow these numbers by each defined area.
No changes for PreventDiscoStorageModif and PreventReconnection as whose behaviours are not affected by zonal rules.

### DIFF
--- a/grid2op/tests/test_RulesByArea.py
+++ b/grid2op/tests/test_RulesByArea.py
@@ -6,7 +6,6 @@
 # SPDX-License-Identifier: MPL-2.0
 # This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
 
-import warnings
 from itertools import chain
 from grid2op.tests.helper_path_test import *
 from grid2op.Exceptions import *
@@ -18,21 +17,10 @@ import warnings
 
 class TestDefaultRulesByArea(unittest.TestCase):
     def setUp(self):
-        params = Parameters()
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore")
-            env_ref = make(
-                "l2rpn_case14_sandbox",
-                test=True,
-                param=params,
-            )
-
-            self.substation_ids = [k for k in range(env_ref.n_sub)]
-            env_ref.close()
-
-            self.rules_1area = RulesByArea([[self.substation_ids]])
-            self.rules_2areas = RulesByArea([[k for k in list_ids] for list_ids in np.array_split(np.random.shuffle(self.substation_ids), 2)])
-            self.rules_3areas = RulesByArea([[k for k in list_ids] for list_ids in np.array_split(np.random.shuffle(self.substation_ids), 3)])
+        n_sub = 14
+        self.rules_1area = RulesByArea([[int(k) for k in range(n_sub)]])
+        self.rules_2areas = RulesByArea([[k for k in np.arange(n_sub,dtype=int)[:8]],[k for k in np.arange(n_sub,dtype=int)[8:]]])
+        self.rules_3areas = RulesByArea([[k for k in np.arange(n_sub,dtype=int)[:4]],[k for k in np.arange(n_sub,dtype=int)[4:9]],[k for k in np.arange(n_sub,dtype=int)[9:]]])
 
     def test_rules_areas(self):
         params = Parameters()
@@ -45,8 +33,9 @@ class TestDefaultRulesByArea(unittest.TestCase):
                         param=params,
                         gamerules_class = rules
                     )
-                lines_by_area = env_ref._game_rules.legal_action.lines_id_by_area
-                line_select = [[int(k) for k in np.random.choice(list_ids, size=3, replace=False)] for list_ids in lines_by_area]
+                self.helper_action = self.env._helper_action_env
+                lines_by_area = self.env._game_rules.legal_action.lines_id_by_area
+                line_select = [[int(k) for k in np.random.choice(list_ids, size=3, replace=False)] for list_ids in lines_by_area.values()]
                 #two lines one sub by area with 1 action in one area per item
                 try:
                     self.env._parameters.MAX_SUB_CHANGED = 1
@@ -54,7 +43,7 @@ class TestDefaultRulesByArea(unittest.TestCase):
                     for line_select_byarea in line_select:
                         act = {
                             "set_line_status": [(LINE_ID, -1) for LINE_ID in line_select_byarea[:2]],
-                            "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in line_select_byarea[0][-1]]}
+                            "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in line_select_byarea[2:]]}
                         }
                         _ = self.helper_action(
                             act,
@@ -71,8 +60,8 @@ class TestDefaultRulesByArea(unittest.TestCase):
                     self.env._parameters.MAX_LINE_STATUS_CHANGED = 1
                     for line_select_byarea in line_select:
                         act = {
-                            "set_line_status": [(LINE_ID, -1) for LINE_ID in line_select_byarea[0]],
-                            "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in line_select_byarea[0][-2:]]}
+                            "set_line_status": [(LINE_ID, -1) for LINE_ID in line_select_byarea[:1]],
+                            "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in line_select_byarea[1:]]}
                         }
                         _ = self.helper_action(
                             act,
@@ -87,8 +76,8 @@ class TestDefaultRulesByArea(unittest.TestCase):
                 self.env._parameters.MAX_SUB_CHANGED = 1
                 self.env._parameters.MAX_LINE_STATUS_CHANGED = 1
                 act = {
-                        "set_line_status": [(LINE_ID, -1) for LINE_ID in list(chain(*[list_ids[0] for list_ids in line_select]))],
-                        "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in list(chain(*[list_ids[1] for list_ids in line_select]))]}
+                        "set_line_status": [(LINE_ID, -1) for LINE_ID in list(chain(*[list_ids[:1] for list_ids in line_select]))],
+                        "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in list(chain(*[list_ids[1:2] for list_ids in line_select]))]}
                 }
                 _ = self.helper_action(
                         act,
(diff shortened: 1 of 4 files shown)


BDonnot (changes requested):

grid2op/Rules/rulesByArea.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        See :func:`BaseRules.__call__` for a definition of the _parameters of this function.
+        """
+        n_sub = env.n_sub
+        n_sub_rule = np.sum([len(set(list_ids)) for list_ids in self.substations_id_by_area.values()])
+        if n_sub_rule != n_sub: 
+            print("The number of listed ids of substations in rule initialization does not match the number of substations of the chosen environement. Look for missing ids or doublon")
+        else:
+            self.lines_id_by_area = {key : sorted(list(chain(*[[item for item in np.where(env.line_or_to_subid == subid)[0]
+                                    ] for subid in subid_list]))) for key,subid_list in self.substations_id_by_area.items()}
```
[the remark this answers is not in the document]
GoubetClem: @BDonnot  you mean add :
* function _initialize_with_env wich does the checking above called only at env creation ? 
*  triggered with a flag like self.initialized_with_env ?

grid2op/Rules/rulesByArea.py
```diff
@@ -0,0 +1,91 @@
...
+    This subclass combine both :class:`LookParam` and :class:`PreventReconnection` to be applied on defined areas of a grid.
+    An action is declared legal if and only if:
+
+      - It doesn't disconnect / reconnect more power lines within each area than  what stated in the actual game _parameters
+        :class:`grid2op.Parameters`
+      - It doesn't attempt to act on more substations within each area that what is stated in the actual game _parameters
+        :class:`grid2op.Parameters`
+      - It doesn't attempt to modify the power produce by a turned off storage unit
+
+    """
```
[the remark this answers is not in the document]
GoubetClem: Just subclass `LookParam` to change and yes I have to add an example

grid2op/Rules/rulesByArea.py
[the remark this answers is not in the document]
GoubetClem: I Think I got it :-)

grid2op/Rules/rulesByArea.py
```diff
@@ -0,0 +1,91 @@
...
+            
+            return PreventReconnection.__call__(self, action, env)
+        
+
+    def can_use_simulate(self, nb_simulate_call_step, nb_simulate_call_episode, param):
+        return LookParam.can_use_simulate(
+            self, nb_simulate_call_step, nb_simulate_call_episode, param
+        )
+    
+    def LookParamByArea(self, action, env):
```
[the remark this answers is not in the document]
GoubetClem: got it :-)

grid2op/tests/test_RulesByArea.py
```diff
@@ -0,0 +1,117 @@
...
+                            "change_bus" : {"lines_or_id":[LINE_ID for LINE_ID in line_select_byarea[2:]]}
+                        }
+                        _ = self.helper_action(
+                            act,
+                            env=self.env,
+                            check_legal=True,
+                        )
+                    raise RuntimeError("This should have thrown an IllegalException")
+                except IllegalAction:
+                    pass
```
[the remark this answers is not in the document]
GoubetClem: got it :-)

grid2op/tests/test_RulesByArea.py
```diff
@@ -0,0 +1,117 @@
...
+import warnings
+
+class TestDefaultRulesByArea(unittest.TestCase):
+    def setUp(self):
+        n_sub = 14
+        self.rules_1area = RulesByArea([[int(k) for k in range(n_sub)]])
+        self.rules_2areas = RulesByArea([[k for k in np.arange(n_sub,dtype=int)[:8]],[k for k in np.arange(n_sub,dtype=int)[8:]]])
+        self.rules_3areas = RulesByArea([[k for k in np.arange(n_sub,dtype=int)[:4]],[k for k in np.arange(n_sub,dtype=int)[4:9]],[k for k in np.arange(n_sub,dtype=int)[9:]]])
+
+    def test_rules_areas(self):
```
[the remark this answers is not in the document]
GoubetClem: question on that "use action that are illegal in one area but you still do action another area"
 I did not understand it was in the scope to pick relevant elementary legal actions within a global action vector when some are illegals. Need advice on how to proceed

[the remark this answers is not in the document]
BDonnot: This is indeed out of scope. 

What I had in mind for this test is to test "when part of an action is legal (say action on zone 2 is perfectly fine) but another part is illegal (say action on zone 1 is illegal) then the whole action should be illegal" [the opposite could happen if the flag is not properly set when checking for legal / illegal)

grid2op/Environment/Environment.py
```diff
@@ -277,6 +277,7 @@ def _init_backend(
         self._check_rules_correct(legalActClass)
                 
         self._game_rules = RulesChecker(legalActClass=legalActClass)
+        self._game_rules.initialize(self)
```
BDonnot, line 280: Something that comes to mind, when the env is copied (init_deepcopy_for_copy or something like this) do we need to do this ?

grid2op/Rules/BaseRules.py
```diff
@@ -20,6 +20,17 @@ class BaseRules(ABC):
...
+    
+    def initialize(self, env):
+        """
+        This function is used to inform the class instance about the environment specification. 
+        It can be the place to assert the defined rules are suited for the environement.
+        Parameters
+        ----------
+        env: :class:`grid2op.Environment.Environment`
+            The environment on which the action is performed. 
+        """
```
BDonnot, line 32: Maybe just say explicitly that the env is not totally initialized when this is called.

grid2op/Rules/RulesChecker.py
```diff
@@ -50,6 +50,17 @@ def __init__(self, legalActClass=AlwaysLegal):
                 warnings.warn("You passed the legal action as an instance that cannot be deepcopied. It will be "
                               "used 'as is', we do not garantee anything if you modify the original object.")
                 self.legal_action = legalActClass
+                
+    def initialize(self, env):
+        """
+        This function is used to inform the class instance about the environment specification. 
+        It can be the place to assert the defined rules are suited for the environement.
+        Parameters
```
BDonnot, line 58: maybe say the env is not fully initialized when this is called

grid2op/Rules/rulesByArea.py
```diff
@@ -0,0 +1,119 @@
...
+    If you want the environment to take into account the rules by area, you can achieve it with:
+
+    .. code-block:
+        import grid2op
+        from grid2op.Rules.rulesByArea import RulesByArea
+        
+        # First you set up the areas within the RulesByArea class
+        my_gamerules_byarea = RulesByArea([[0,1,2,3,4,5,6,7],[8,9,10,11,12,13,14]])
+        # Then you create your environment with it:
+        NAME_OF_THE_ENVIRONMENT = "rte_case14_realistic"
```
BDonnot, line 42: can you put "l2rpn_case14_sandbox" instead of "rte_case14_realistic" as the example name ? (I try to remove any reference non l2rpn environments)

grid2op/Rules/rulesByArea.py
```diff
@@ -0,0 +1,119 @@
...
+        Parameters
+        ----------
+        env: :class:`grid2op.Environment.Environment`
+            An environment instance properly initialized.
+
+        """
+        n_sub = env.n_sub
+        n_sub_rule = np.sum([len(set(list_ids)) for list_ids in self.substations_id_by_area.values()])
+        if n_sub_rule != n_sub: 
+            print("The number of listed ids of substations in rule initialization does not match the number of substations of the chosen environement. Look for missing ids or doublon")
```
BDonnot, line 69: can you raise a Grid2opException here instead ?

grid2op/Rules/rulesByArea.py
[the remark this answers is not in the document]
BDonnot: I think you got it too thanks :-)

grid2op/Rules/rulesByArea.py
```diff
@@ -0,0 +1,119 @@
...
+            ids = [[k for k in np.where(aff_subs)[0] if k in sub_ids] for sub_ids in self.substations_id_by_area.values()]
+            return False, IllegalAction(
+                "More than {} substation affected by the action in one area: {}"
+                "".format(env.parameters.MAX_SUB_CHANGED, ids)
+            )
+        return True, None
+
+
+
+
```
BDonnot, line 119: just one blank line at the end of the file :-)

grid2op/tests/test_RulesByArea.py
```diff
@@ -0,0 +1,146 @@
+# Copyright (c) 2023, RTE (https://www.rte-france.com)
+# See AUTHORS.txt
```
BDonnot, line 2: I just noticed it here, did you add yourself in the AUTHORS.txt ?

BDonnot (changes requested):
A few minor changes and a test missing but beside that everything looks good, thanks :-)

grid2op/Environment/Environment.py
GoubetClem: Mirror question : is _game_rules class attribute is copied as a class or an instance with _custom_deepcopy_for_copy(self) ? Seems to be copy.deepcopy which supposedly copy each attribute of the gamerules, thus even the created ones when initialized. Can you confirm the thought ?

grid2op/Rules/RulesChecker.py
GoubetClem: ok !

grid2op/Rules/rulesByArea.py
GoubetClem: ok !

grid2op/tests/test_RulesByArea.py
GoubetClem: Nope, thanks for notifying !

grid2op/Rules/rulesByArea.py
GoubetClem: ok !

grid2op/Rules/BaseRules.py
GoubetClem: ok !

grid2op/Environment/Environment.py
BDonnot: Good question. the "deep copy" does not really work (yet) with grid2op.

The "_custom_deepcopy_for_copy" is here precisely because the deepcopy does not work. So I would not draw any conclusion from the behaviour of deepcopy to "_custom_deepcopy_for_copy". What does the implementation says about the `gamerules` ?

PR #482, opened 2023-07-06
Assistant baseline

marota:


grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+from grid2op.Agent.recoPowerlineAgent import RecoPowerlineAgent
+import pandas as pd
+
+#try:
+#    from lightsim2grid import LightSimBackend
+#
+#    bkclass = LightSimBackend
+#    # raise ImportError()
+#except ImportError as excq_:
+#    from grid2op.Backend import PandaPowerBackend
```
BDonnot, line 18: Needs to be removed

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2019-2020, RTE (https://www.rte-france.com)
```
BDonnot, line 1: put 2023 instead of 2020 here :-)

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+#    from lightsim2grid import LightSimBackend
+#
+#    bkclass = LightSimBackend
+#    # raise ImportError()
+#except ImportError as excq_:
+#    from grid2op.Backend import PandaPowerBackend
+
+
+
+class AlertAgent(RecoPowerlineAgent):
```
BDonnot, line 22: I think it would be better not to inherit from RecoPowerlineAgent but rather to pass a "controler" (which by default would be a RecoPowerlineAgent) in the `__init__`.
The "controler" would handle the actions on the grid, and this "AlertAgent" would handle the alerts.

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+    This is a :class:`AlertAgent` example, which will attempt to reconnect powerlines and send alerts on the worst possible attacks: for each disconnected powerline
+    that can be reconnected, it will simulate the effect of reconnecting it. And reconnect the one that lead to the
+    highest simulated reward. It will also simulate the effect of having a line disconnection on attackable lines and raise alerts for the worst ones
+
+    """
+
+    def __init__(self, action_space,percentage_alert=30,simu_step=0):
+        RecoPowerlineAgent.__init__(self, action_space)
+        self.percentage_alert = percentage_alert
+        self.simu_step=simu_step
```
BDonnot, line 33: We would have:

```python
def __init__(self, action_space, controler=RecoPowerlineAgent, percentage_alert=30, simu_step=0):
   if isinstance(controler, type):
       # controler is passed as a class, and not as an object
       self.controler = controler(action_space)
    else:
        self.controler = controler
    # rest of your implementation 
```

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def act(self, observation, reward, done=False):
+        action=super().act(observation, reward, done=False)
```
BDonnot, line 36: have here:
```python
action = self.controler.act(observation, reward, done)
``` 
instead

BDonnot: (and if you want to keep it this way, it should be `done=done` and not `done=False` here )

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+        alertable_line_ids=observation.alertable_line_ids
+
+        #simu d'analyse de sécurité à chaque pas de temps sur les lignes attaquées
+        n_alertable_lines=len(alertable_line_ids)
+        nb_overloads=np.zeros(n_alertable_lines)
+        rho_max_N_1=np.zeros(n_alertable_lines)
```
BDonnot, line 42: this could be computed once and for all in the `reset` for example (or in the `__init__`)

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # test which backend to know which method to call
+        N_1_actions=[self.action_space({"set_line_status": [(id_, -1)]}) for id_ in alertable_line_ids]
```
BDonnot, line 45: same, why building it each time ? Build it in `reset` or in `__init__`

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+        for i, action_to_simulate in enumerate(N_1_actions):
+
+            #check that line is not already disconnected
+            if (observation.line_status[alertable_line_ids[i]]):
+                (
+                    simul_obs,
+                    simul_reward,
+                    simul_has_error,
+                    simul_info,
+                ) = observation.simulate(action_to_simulate,time_step=self.simu_step)
```
BDonnot, line 55: what do you do if your action is illegal ? It will then be equivalent to a game over ?
Beside, why not consider the action of the `controler` here in this simulate (so simulating `action_of_the_actor + powerline_disconnection` ? This would be more interesting I believe

grid2op/Agent/alertAgent.py
```diff
@@ -0,0 +1,69 @@
...
+
+                rho_simu=simul_obs.rho
+                if(not simul_has_error):
+                    nb_overloads[i]=np.sum(rho_simu >= 1)
+                    rho_max_N_1[i]=np.max(rho_simu)
+
+        df_to_sort=pd.DataFrame({"nb_overloads":nb_overloads,"rho_max_N_1":rho_max_N_1})
+        indices=df_to_sort.sort_values(['nb_overloads','rho_max_N_1'], ascending=False).index
```
BDonnot, line 63: why not staying in numpy world ? This is un necessary to convert to pandas, then sort. You can sort with numpy and retrieve the index directly.

grid2op/data_test/l2rpn_idf_2023_with_alert/config.py
```diff
@@ -9,6 +9,8 @@
 try:
     from grid2op.l2rpn_utils import ActionIDF2023, ObservationIDF2023
 except ImportError:
+    from grid2op.Observation import CompleteObservation
+    import warnings
```
BDonnot, line 13: Nice catch, can you check if it's correctly everywhere ? In the input_data_* of the starting kit and in the training environment of l2rpn_idf_2023 ?

grid2op/tests/test_baseline_alert.py
```diff
@@ -0,0 +1,47 @@
...
+            runner = Runner(**env.get_params_for_runner(), agentClass=None ,agentInstance=my_agent)
+
+            res = runner.run(nb_episode=1, nb_process=1, path_save=None,agent_seeds=[0],env_seeds=[0],max_iter=3,
+                             add_detailed_output=True)
+            id_chron, name_chron, cum_reward, nb_time_step, max_ts, episode_data = res[0]
+
+            # test if the number of alerts sent on lines are recovered
+            alerts_count =np.sum([obs.active_alert for obs in episode_data.observations[1:]]
+                                  ,axis=0)
+            print(alerts_count)
```
BDonnot, line 36: no print in the test please

grid2op/tests/test_baseline_alert.py
```diff
@@ -0,0 +1,47 @@
...
+            assert(np.all(alerts_count==[3, 3, 2, 0, 1, 0, 0, 0, 0, 0]))
+
+            # test that we observe the expected alert rate
+            nb_alertable_lines =len(env.alertable_line_names)
+            ratio_alerts_step =np.sum(alerts_count ) /(nb_time_step*nb_alertable_lines)
+            assert(np.round(ratio_alerts_step ,decimals=1 )==np.round(percentage_alert/100 ,decimals=1))
+
+            #check that alert agent is not doing any intervention on the grid in this short time frame
+            #as the reco power line, it should only do actions to reconnect lines when allowed, but cannot happen in this short time frame
```
BDonnot, line 45: and what happens if precisely the "controler" acts on the grid ? Is its action "cut" and limited to the alert in this case ? This should be tested too

BDonnot (changes requested):
Some changes need to be made.

I think the inheritance is not the correct tools for the AlertAgent, it should rather have an "actor" member and deleguate to the actor the action on the grid itself. And then take of what alerts should be made.

The "simulate" logic does not catch the "illegal action" or "ambiguous" action.

Some tests are missing, in the case the "controler" performs an action for example.

BDonnot (approved):
Looks good if tests pass

PR #492, opened 2023-07-19
Dev 1.9.1 l2rpnidf2023 score with alert minimal version

marota:
New PR starting from a verified commit with tests that passed.
Inserting more up-to date code of alertTrustScore. But with code to compute new kpis commented to make a minimal working version


marota:
@BDonnot good for review in my opinion

laure-crochepierre:
@marota you must switch your PR to branch [rte-france:dev_1.9.2](https://github.com/rte-france/Grid2Op/tree/dev_1.9.2)

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
+
+import numpy as np
+from grid2op.Reward import AlertReward
+from grid2op.dtypes import dt_float
+
+SURVIVOR_TIMESTEPS = 12 #nb timesteps to be considered survivor of an attack
```
BDonnot, line 13: no global variables please. This is customized in the parameters of the environment.

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+    #             reward_max_blackout=0.0,
+    #             reward_end_episode_bonus=0.0,
+    #             min_score=-3):
+    def __init__(self,
+                 logger=None,
+                 reward_min_no_blackout=-1.0,
+                 reward_min_blackout=-10.0,
+                 reward_max_no_blackout=1.0,
+                 reward_max_blackout=2.0,
+                 reward_end_episode_bonus=1.0,
```
BDonnot, line 50: can we put by default the parameters that are used for the competition ? Thanks

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+        #self.no_alert_attack_blackout = 0
+        #self.blackout_encountered = False
+            
+    def __call__(self, action, env, has_error, is_done, is_illegal, is_ambiguous):
+
+        self.blackout_encountered = self.is_in_blackout(has_error, is_done)
+        
+        score_ep = 0.
+        if self._is_simul_env:
+            return score_ep
```
BDonnot, line 102: Do this before anything else (faster)

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+            #    lines_attacked_no_blackout = env._time_since_last_attack > 0
+            #
+            #    self.alert_attack_no_blackout += np.sum(lines_alerted_beforeattack[lines_attacked_no_blackout])
+            #    self.no_alert_attack_no_blackout += np.sum(~lines_alerted_beforeattack[lines_attacked_no_blackout])
+            
+            return score_ep
+        
+    @staticmethod
+    def _normalisation_fun(cm_reward, cm_reward_min_ep, cm_reward_max_ep,min_score,max_score):
+        standardized_score = np.round((cm_reward - cm_reward_min_ep) / (cm_reward_max_ep - cm_reward_min_ep +1e-5),4)
```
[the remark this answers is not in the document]
BDonnot: Why rounding ? It should not be rounded at this stage. If a user wants to round it it's his responsibility, not the one of the class.

grid2op/tests/test_RewardAlertCostScore.py
```diff
@@ -0,0 +1,420 @@
...
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
+
+import warnings
+import numpy as np
+import unittest
+import tempfile
+import grid2op
+from grid2op.Reward import _AlertCostScore, _AlertTrustScore
```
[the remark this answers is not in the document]
BDonnot: (i don't read more at the moment)

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
...
+def _get_steps_attack(kwargs_opponent, multi=False):
+    """computes the steps for which there will be attacks"""
+    ts_attack = np.array(kwargs_opponent["steps_attack"])
+    res = []
+    for i, ts in enumerate(ts_attack):
+        if not multi:
+            res.append(ts + np.arange(kwargs_opponent["duration"]))
+        else:
+            res.append(ts + np.arange(kwargs_opponent["duration"][i]))
+    return np.unique(np.concatenate(res).flatten())
```
BDonnot, line 67: no copy paste!

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
...
+
+    def attack(self, observation, agent_action, env_action, budget, previous_fails): 
+        if observation is None:
+            return None, None
+        current_step = self.env.nb_time_step
+        if current_step not in self.steps_attack: 
+            return None, None
+        
+        return self.custom_attack, self.duration
+
```
BDonnot, line 95: no copy paste

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
...
+            return None, None
+
+        current_step = self.env.nb_time_step
+        if current_step not in self.steps_attack: 
+            return None, None
+        
+        index = self.steps_attack.index(current_step)
+
+        return self.custom_attack[index], self.duration[index]
+
```
BDonnot, line 125: no copy paste

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
...
+
+# Test alert blackout / tets alert no blackout
+class TestAlertTrustScoreNoBlackout(unittest.TestCase):
+    """test the basic bahavior of the assistant alert feature when no attack occur """
+
+    def setUp(self) -> None:
+        self.env_nm = os.path.join(
+            PATH_DATA_TEST, "l2rpn_idf_2023_with_alert"
```
BDonnot, line 133: why not using the "l2rpn_idf_2023" environment ?

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
+# Copyright (c) 2019-2023, RTE (https://www.rte-france.com)
+# See AUTHORS.txt
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
```
BDonnot, line 7: I don't understand what is done here. Is this all copy paste from the AlertReward ? 

If so what's the point, as everything is tested in AlertReward already ? What is the difference between the _AlertTrustScore and the AlertReward.

These are the only things that needs to be tested. Making copy paste of previous tests is not testing more stuff.

BDonnot: I stop my reading here

grid2op/tests/test_score_idf_2023.py
```diff
@@ -49,7 +51,9 @@ def act(self, observation: BaseObservation, reward: float, done: bool = False) -
 class TestScoreL2RPN2023(unittest.TestCase):
     
     def setUp(self) -> None:
-        env_name = "l2rpn_case14_sandbox"
+        env_name = os.path.join(
+            PATH_DATA_TEST, "l2rpn_idf_2023_with_alert"
```
BDonnot, line 55: please change it to "l2rpn_idf_2023"

grid2op/tests/test_score_idf_2023.py
```diff
@@ -71,7 +75,7 @@ def setUp(self) -> None:
             params.LIMIT_INFEASIBLE_CURTAILMENT_STORAGE_ACTION = True
         self.seed = 0
         self.scen_id = 0
-        self.nb_scenario = 2
```
BDonnot, line 74: and if you keep the "l2rpn_idf_2023" magically you can keep the 2 scenarios and all the assert that were there in the first place.

grid2op/tests/test_score_idf_2023.py
```diff
@@ -107,45 +111,65 @@ def test_score_helper(self):
...
+                assert nres_score == 100.
+                assert ep_score == 0.6 * op_score + 0.15 * nres_score + 0.25 * assistant_score
+                assert assistant_score == 100. #no blackout with no disconnections
+                assert op_score == 0
+            
+                
+        finally:
+            my_score.clear_all()
 
     def test_min_score(self):
```
BDonnot, line 174: the `min_assistant_cost_score` is added in the class that is tested, but is never tested. Maybe change the name of this test, and test the same kind of thins with the assistant (making sure that the score cannot go below the min)

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -72,25 +73,25 @@ class ScoreL2RPN2023(ScoreL2RPN2020):
...
+            nb_process_stats=1,
+            scale_assistant_score=100.0,
+            scale_nres_score=100.,
+            weight_op_score=0.6,
+            weight_assistant_score=0.25,
+            weight_nres_score=0.15,
+            # weight_confidence_assistant_score=0.7,
+            min_nres_score=-100,
+            min_assistant_cost_score=-100,
+            add_nb_highres_sim=False,
```
BDonnot, line 93: this argument is not at the right place. It should be right after `score_names` to follow the order of ScoreL2RPN2020

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -103,32 +104,41 @@ def __init__(
             nb_process_stats=nb_process_stats,
             scores_func={
                 "grid_operational_cost": L2RPNSandBoxScore,
+                "assistant_confidence": _AlertTrustScore,
+                # "assistant_cost": _AlertCostScore,
```
BDonnot, line 108: remove as not used

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -103,32 +104,41 @@ def __init__(
...
-        assert(weight_op_score + weight_assistant_score + weight_nres_score==1.)
-        assert(all([weight_confidence_assistant_score>=0., weight_confidence_assistant_score<=1.]))
-        
+
+        test_weights = weight_op_score + weight_assistant_score + weight_nres_score
+        if test_weights != 1.0:
+            raise Grid2OpException(
+                'The weights of each component of the score shall sum to 1'
+            )
+        # assert(all([weight_confidence_assistant_score>=0., weight_confidence_assistant_score<=1.]))
```
BDonnot, line 123: why is this assert removed ?

BDonnot (changes requested):
Some minor things to do, but also:

- even in the test, avoid copy paste, I did not understand what was tested and what was not
- make clearer in the name of the tests what is tested
- why the formula in the "AlertReward" is modified ? I don't follow, we extensively tested it with the previous formula. What's the difference ? Why changing things silently at the last minute ?

grid2op/Reward/_alertTrustScore.py
marota: I agree, it was there, I will remove it

grid2op/Reward/_alertTrustScore.py
marota: ok

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+        #self.no_alert_attack_blackout = 0
+        #self.blackout_encountered = False
+        return super().initialize(env)
+            
+    def reset(self, env):
+        super().reset(env)
+        self.cumulated_reward = 0
+        #KPIs
+        self.total_nb_attacks = 0
+        self.nb_last_attacks=0
```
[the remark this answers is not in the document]
marota: ok

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+        
+        score_ep = 0.
+        if self._is_simul_env:
+            return score_ep
+            
+        res = super().__call__(action, env, has_error, is_done, is_illegal, is_ambiguous)
+        self.cumulated_reward += res
+
+        lines_attacked = env._time_since_last_attack == 0
+        self.total_nb_attacks += np.sum(lines_attacked)
```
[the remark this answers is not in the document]
marota: ok

grid2op/utils/l2rpn_idf_2023_scores.py
marota: You mean add_nb_highres_sim ?
In that case it will look something like that( after nb_process_stats actually)
image

Do you want me to add scores_func and score_names argument also ? (They are missing in ScoreL2RPN2023, only hardcoded after

grid2op/utils/l2rpn_idf_2023_scores.py
marota: I did not chage it, @laure-crochepierre  @GoubetClem any idea ?

I would say probably because we only use one sub-score for the Assistant Score and that this weight is not necessary anymore

grid2op/utils/l2rpn_idf_2023_scores.py
marota: ok

grid2op/Reward/_alertTrustScore.py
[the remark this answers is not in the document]
marota: Because then you get score of 0.99993.
Where should we round this then ? Here it is to compensate for the addition of the tolerance that is artificial

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+            #
+            #    self.alert_attack_no_blackout += np.sum(lines_alerted_beforeattack[lines_attacked_no_blackout])
+            #    self.no_alert_attack_no_blackout += np.sum(~lines_alerted_beforeattack[lines_attacked_no_blackout])
+            
+            return score_ep
+        
+    @staticmethod
+    def _normalisation_fun(cm_reward, cm_reward_min_ep, cm_reward_max_ep,min_score,max_score):
+        standardized_score = np.round((cm_reward - cm_reward_min_ep) / (cm_reward_max_ep - cm_reward_min_ep +1e-5),4)
+        #in case cm_reward_min_ep=cm_reward_max_ep=0, score is 0.0
```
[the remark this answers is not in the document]
marota: You meant this line ?

score_ep = (cm_reward_min_ep!=cm_reward_max_ep)*min_score + (max_score - min_score)

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+        elif(self.blackout_encountered):
+            if(nb_last_attacks==0):
+                cm_reward_min_ep = lambda k: self.reward_min_no_blackout * k
+                cm_reward_max_ep = lambda k: self.reward_max_no_blackout * k
+            elif(nb_last_attacks>=1):
+                cm_reward_min_ep = lambda k: self.reward_min_no_blackout * (k - nb_last_attacks) + self.reward_min_blackout
+                cm_reward_max_ep = lambda k: self.reward_max_no_blackout * (k - nb_last_attacks) + self.reward_max_blackout
+        else:
+            cm_reward_min_ep = lambda k: self.reward_min_no_blackout * k
+            cm_reward_max_ep = lambda k: self.reward_max_no_blackout * k + self.reward_end_episode_bonus
```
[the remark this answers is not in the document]
marota: That was how it was done, perhaps some reason...
But indeed just assignig the value can work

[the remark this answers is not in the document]
marota: I did it

grid2op/Reward/alertReward.py
```diff
@@ -196,7 +206,15 @@ def __call__(self, action, env, has_error, is_done, is_illegal, is_ambiguous):
...
                 # update the state of the environment
                 env._was_alert_used_after_attack[lines_attack] = 1 - alert_send * 2
-                res = (self.reward_min_no_blackout - self.reward_max_no_blackout) * np.mean(alert_send) + self.reward_max_no_blackout
+
+                res=0
+                for alert in alert_send:
+                    if(alert):
+                        res += self.reward_min_no_blackout
+                    else:
+                        res += self.reward_max_no_blackout
```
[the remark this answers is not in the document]
marota: "why the formula in the "AlertReward" is modified ? I don't follow, we extensively tested it with the previous formula. What's the difference ? Why changing things silently at the last minute ?"

Quand 2 attaques sont simultanées sans blackout ont fait la moyenne des reward:

    Ceci me semble effectivement à faire en cas de blackout car on ne sait pas quelle alerte a été pertinente
    Mais sans blackout, on sait que les 2 alertes étaient pertinentes. Il me semble qu'il faut les cumuler de fait. de manière analogue a si les attaques arrivaient juste l'une après l'autre (pour lesquels on aura en reward cumulée la somme des 2). Chaque réponse de l'agent compte alors

Ceci pose un gros problème dans le calcul des min/max score de _alert_trust_score:

    dans ce cas d'attaque simultanée avec moyenne des rewards, il ne faut pas compter 2 mais 1 attaque pour calculer cm_reward_min_ep et cm_reward_max_ep

Donc compter le nombre d'attaque simultanées au cours du scénarios, ce qui est étrange et pas trivial

Les tests passent toujours par ailleurs

grid2op/tests/test_RewardAlertCostScore.py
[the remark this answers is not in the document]
marota: @laure-crochepierre  can you explain ?

grid2op/tests/test_alert_feature.py
```diff
@@ -254,6 +259,33 @@ def _aux_obs_init(self, obs):
...
+
+        assert (obs2.time_since_last_alert == np.array([0, -1, -1, -1, -1, -1, -1, -1, -1, -1])).all()
+
+        obs2bis, reward, done, info = self.env.step(self.env.action_space({"raise_alert": [1]}))
+        assert (obs2bis.time_since_last_alert == np.array([1, 0, -1, -1, -1, -1, -1, -1, -1, -1])).all()
+
+        obs3 : BaseObservation = self.env.reset()
+        assert (obs3.time_since_last_alert == obs1.time_since_last_alert).all()
+
+    def test_reset_reward(self) -> None :
```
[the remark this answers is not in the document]
marota: Ok I haven't done it myseff either. Do you suggest something else ?

grid2op/tests/test_alert_score.py
```diff
@@ -40,6 +40,11 @@
 
 ATTACKED_LINE = "48_50_136"
 
+DEFAULT_ALERT_REWARD_PARAMS = dict(reward_min_no_blackout=-1.0,
+                                   reward_min_blackout=-10.0, 
+                                   reward_max_no_blackout=1.0,
+                                   reward_max_blackout=2.0,
+                                   reward_end_episode_bonus=42.0)
 
 def _get_steps_attack(kwargs_opponent, multi=False):
```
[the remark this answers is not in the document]
marota: this is to test the Alert_Trust_Score here, see _AlertTrustScore(**DEFAULT_PARAMS_TRUSTSCORE) in make.

This was done here for a configurable test if you want to change parameters. Good idea from @GoubetClem  or @laure-crochepierre

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
...
+import os
+import tempfile
+from grid2op.Observation import BaseObservation
+from grid2op.tests.helper_path_test import *
+
+from grid2op import make
+from grid2op.Reward import _AlertTrustScore
+from grid2op.Parameters import Parameters
+from grid2op.Exceptions import Grid2OpException
+from grid2op.Runner import Runner  # TODO
```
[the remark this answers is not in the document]
marota: Probably a TODO

grid2op/tests/test_alert_trust_score.py
marota: from where is the copy paste ? this was like this but I can change

grid2op/tests/test_alert_trust_score.py
marota: It is apparently an attackant sending custom attacks.
What should be done in that case ?

grid2op/tests/test_alert_trust_score.py
marota: Same as above

grid2op/tests/test_alert_trust_score.py
marota: This is to test _AlertTrustScore.

I don't understand what is misleading you here.

What is tested is the cumulated reward, the min and max possible cumulated reward, the score. This was the format of test chosen before hand

marota:
By the way @BDonnot there is a new test in test_score_idf_2023 from Clement or Laure that was unning properly before my merge with you latests commits and is failing know, this is very starnge:

Traceback (most recent call last):
  File "/Grid2Op/grid2op/tests/test_score_idf_2023.py", line 233, in test_spec
    assert abs(nres_score + 100.) <= tol
    
 What I observe is that the nres_score is always a 100 whatever the level of curtailment you do with the agent.
 I have a hard time finding why, if it comes from the actions not running properly or some statistics that have changed, or...

grid2op/Reward/_alertTrustScore.py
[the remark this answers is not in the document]
marota: I did this

        if(cm_reward_min_ep==cm_reward_max_ep):
            score_ep = 0.
        else:
            score_ep = min_score + (max_score - min_score) * standardized_score

grid2op/Reward/alertReward.py
```diff
@@ -115,6 +116,15 @@ def initialize(self, env: "grid2op.Environment.BaseEnv"):
         
         self._i_am_simulate = self.is_simulated_env(env)
```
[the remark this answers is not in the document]
marota: @laure-crochepierre you confirm ?

grid2op/Reward/_alertTrustScore.py
marota: You mean in the __call__ method ?

Like this before the blackout_encountered line ?
image

BDonnot:
For the conflicts, I cherry picked Laure's changes to alertReward and test_alert_feature. I would prefer no to see any change to either of this file (except of course if there is a bug spotted)

grid2op/utils/l2rpn_idf_2023_scores.py
laure-crochepierre: yes, I think you can undo it. It looks like a typo. It was just modified in the last commit [synchro merge with branch 1.9.2](https://github.com/rte-france/Grid2Op/pull/492/commits/984c01ee6ba413ed6296b6b9c5951d1894ae9258)

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -152,33 +162,34 @@ def _compute_episode_score(
...
         # assistant_cost_score = float(other_rewards[-1][key_score_file])
-        assistant_cost_score  = 0 #self.scale_assistant_score * assistant_cost_score
-        
-        assistant_score = self.weight_confidence_assistant_score * assistant_confidence_score +\
-             (1. - self.weight_confidence_assistant_score) * assistant_cost_score
+        # assistant_cost_score = max(assistant_cost_score, self.min_assistant_cost_score / self.scale_assistant_score)
+        # assistant_cost_score  = self.scale_assistant_score * assistant_cost_score
+
+        # assistant_score = self.weight_confidence_assistant_score * assistant_confidence_score +\
+        #      (1. - self.weight_confidence_assistant_score) * assistant_cost_score
```
laure-crochepierre, line 190: A lot of this has been commented, we should maybe juste remove it (we no longer have two components in the assistant score).

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -103,32 +104,41 @@ def __init__(
...
+            )
+        # assert(all([weight_confidence_assistant_score>=0., weight_confidence_assistant_score<=1.]))
+
         self.scale_assistant_score = scale_assistant_score
         self.scale_nres_score = scale_nres_score
         self.weight_op_score = weight_op_score
         self.weight_assistant_score = weight_assistant_score
         self.weight_nres_score = weight_nres_score
-        self.weight_confidence_assistant_score = weight_confidence_assistant_score
+        # self.weight_confidence_assistant_score = weight_confidence_assistant_score
```
laure-crochepierre, line 130: This has been commented, we should maybe juste remove it (we no longer have two components in the assistant score).

marota:
> > By the way @BDonnot there is a new test in test_score_idf_2023 from Clement or Laure that was unning properly before my merge with you latests commits. It is failing know, this is very strange:
> > Traceback (most recent call last): File "/Grid2Op/grid2op/tests/test_score_idf_2023.py", line 233, in test_spec assert abs(nres_score + 100.) <= tol
> > here and after: https://github.com/marota/Grid2Op/blob/984c01ee6ba413ed6296b6b9c5951d1894ae9258/grid2op/tests/test_score_idf_2023.py#L225
> > What I observe is that the nres_score is always a 100 whatever the level of curtailment you do with the agent. I have a hard time finding why, if it comes from the actions not running properly or some statistics that have changed, or...
> 
> This test exists on the 1.9.1 version and on the dev_1.9.2 version and it passes. I don't know what is modified, but maybe it's linked with the fact that instead of using the `env_name = "l2rpn_case14_sandbox"` in this class, the `l2rpn_idf_2023_with_alert` is used. And... Maybe, but just maybe, this test won't pass with this environment because, for example, there is no renewable energy there.
> 
> Test passes on ci for 1.9.1 and dev_1.9.2 it should pass there too.

Yes indeed this was another test which was added. you are right, I made a shortcut.
This test was passing in https://github.com/rte-france/Grid2Op/pull/492/commits/5c389a3e5089517f45cd87c99f3223b0fd5661a5 with no changes, so I would say the problem we see is not because of this change of grid2Op environment which happened earlier (not me, so some time ago)

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -103,32 +104,41 @@ def __init__(
             nb_process_stats=nb_process_stats,
             scores_func={
                 "grid_operational_cost": L2RPNSandBoxScore,
+                "assistant_confidence": _AlertTrustScore,
+                # "assistant_cost": _AlertCostScore,
                 "new_renewable_sources_usage": _NewRenewableSourcesUsageScore,
             },
             score_names=["grid_operational_cost_scores",
+                         "assistant_confidence_scores",
+                         # "assistant_cost_scores",
```
laure-crochepierre, line 113: This has been commented, we should maybe juste remove it (we no longer have two components in the assistant score).

grid2op/utils/l2rpn_idf_2023_scores.py
laure-crochepierre, line 108: This has been commented, we should maybe juste remove it (we no longer have two components in the assistant score).

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -103,32 +104,41 @@ def __init__(
             nb_process_stats=nb_process_stats,
             scores_func={
                 "grid_operational_cost": L2RPNSandBoxScore,
+                "assistant_confidence": _AlertTrustScore,
```
laure-crochepierre, line 107: shouldn't "confidence" be renamed to "trust"?

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -103,32 +104,41 @@ def __init__(
...
+                # "assistant_cost": _AlertCostScore,
                 "new_renewable_sources_usage": _NewRenewableSourcesUsageScore,
             },
             score_names=["grid_operational_cost_scores",
+                         "assistant_confidence_scores",
```
laure-crochepierre, line 112: shouldn't "confidence" be renamed to "trust"?

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -72,25 +73,25 @@ class ScoreL2RPN2023(ScoreL2RPN2020):
...
+            min_losses_ratio=0.8,
+            verbose=0,
+            max_step=-1,
+            nb_process_stats=1,
+            scale_assistant_score=100.0,
+            scale_nres_score=100.,
+            weight_op_score=0.6,
+            weight_assistant_score=0.25,
+            weight_nres_score=0.15,
+            # weight_confidence_assistant_score=0.7,
```
laure-crochepierre, line 90: This has been commented, we should maybe juste remove it (we no longer have two components in the assistant score).

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -7,8 +7,9 @@
 # This file is part of Grid2Op, Grid2Op a testbed platform to model sequential decision making in power systems.
 
 from grid2op.utils.l2rpn_2020_scores import ScoreL2RPN2020
-from grid2op.Reward import L2RPNSandBoxScore, _NewRenewableSourcesUsageScore, _AssistantConfidenceScore, _AssistantCostScore
+from grid2op.Reward import L2RPNSandBoxScore, _NewRenewableSourcesUsageScore, _AlertTrustScore  # , _AlertCostScore
```
laure-crochepierre, line 10: This has been commented, we should maybe juste remove it (we no longer have two components in the assistant score).

grid2op/tests/test_score_idf_2023.py
laure-crochepierre: Yes I did this (if I remember well), going back to the env "l2rpn_idf_2023" should fix it + going back to the previous values in the test.

grid2op/tests/test_alert_trust_score.py
laure-crochepierre: The point was to test that all cases covered by  tests on the AlertReward are covered in the _AlertTrustScore. 
The attributes that are tested must be different though

grid2op/tests/test_alert_score.py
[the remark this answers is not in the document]
laure-crochepierre: Yes the name is misleading, as the file had this name before the scoring was developped. This file test *AlertReward* and the file name must be changed. 

Alert_Trust_Score is tested in file "test_alert_trust_score.py"

grid2op/tests/test_alert_feature.py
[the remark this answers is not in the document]
laure-crochepierre: As discussed,  test_reset_reward is now testing the reward reset.

grid2op/Reward/alertReward.py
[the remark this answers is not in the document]
laure-crochepierre: As discussed with @BDonnot its done.

grid2op/tests/test_RewardAlertCostScore.py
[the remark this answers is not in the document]
laure-crochepierre: I have no explanation, as I didn't code in this file. It should *only* test  _AlertCostScore behaviour. Tests for _AlertTrustScore must be moved to the other test file test_alert_trust_score

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,168 @@
...
+                 reward_max_blackout,
+                 reward_end_episode_bonus)
+        
+        self.min_score = dt_float(min_score)
+        self.max_score = dt_float(1.0)
+        
+    def initialize(self, env):
+        self._is_simul_env = self.is_simulated_env(env)
+        #SURVIVOR_TIMESTEPS=env._al
+
```
laure-crochepierre, line 66: this comment should be removed

grid2op/Reward/_alertTrustScore.py
laure-crochepierre: yes I think so

marota:
> > > By the way @BDonnot there is a new test in test_score_idf_2023 from Clement or Laure that was unning properly before my merge with you latests commits. It is failing know, this is very strange:
> > > Traceback (most recent call last): File "/Grid2Op/grid2op/tests/test_score_idf_2023.py", line 233, in test_spec assert abs(nres_score + 100.) <= tol
> > > here and after: https://github.com/marota/Grid2Op/blob/984c01ee6ba413ed6296b6b9c5951d1894ae9258/grid2op/tests/test_score_idf_2023.py#L225
> > > What I observe is that the nres_score is always a 100 whatever the level of curtailment you do with the agent. I have a hard time finding why, if it comes from the actions not running properly or some statistics that have changed, or...
> > 
> > 
> > This test exists on the 1.9.1 version and on the dev_1.9.2 version and it passes. I don't know what is modified, but maybe it's linked with the fact that instead of using the `env_name = "l2rpn_case14_sandbox"` in this class, the `l2rpn_idf_2023_with_alert` is used. And... Maybe, but just maybe, this test won't pass with this environment because, for example, there is no renewable energy there.
> > Test passes on ci for 1.9.1 and dev_1.9.2 it should pass there too.
> 
> Yes indeed this was another test which was added. you are right, I made a shortcut. This test was passing in [5c389a3](https://github.com/rte-france/Grid2Op/commit/5c389a3e5089517f45cd87c99f3223b0fd5661a5) with no changes, so I would say the problem we see is not because of this change of grid2Op environment which happened earlier (some time ago with Clement).
> 
> Either we wait for his come back @GoubetClem , or we choose to use l2rpn_case14_sandbox ?
> 
> Actually now I understand what you mean from this change: [grid2op/Reward/_newRenewableSourcesUsageScore.py](https://github.com/rte-france/Grid2Op/commit/c2f4640853ca466a6c1237c904b76998d2e3a1b3)

The explanation @BDonnot is because 50% of curtailment does not work apparently

InvalidRedispatching('\nThis is an attempt to explain why the dispatch did not succeed and caused a game over.\nTo compensate the increase of loads and / or decrease of renewable energy (due to naturl causes but also through curtailment) and / or variation in the storage units, the generators should increase their total production of 49.35MW (in total).\nBut, if you take into account the generator constraints (pmax and max_ramp_up) you can have at most 43.90MW.\nIndeed at time t, generators are in state:\n\t[  0.    0.  134.4   0.    0.    0.    0.  399.    0.   27.4]\ntheir ramp max is:\n\t[ 1.4  1.4 10.4  1.4  2.8  2.8  4.3  2.8  8.5  9.9]\n and pmax is:\n\t[ 50.  50. 250.  50. 100. 100. 150. 400. 300. 350.]\nWrapping up, each generator can increase at maximum of:\n\t[ 1.4  1.4 10.4  1.4  2.8  2.8  4.3  1.   8.5  9.9]\nNB: if you did not do any dispatch during this episode, it would have been possible to meet these constraints. This situation is caused by not having enough degree of freedom to "compensate" the variation of the load due to (most likely) an "over usage" of redispatching feature (some generators stuck at pmax as a consequence of your redispatching. They can\'t increase their productions to meet the increase in demand or decrease of renewables)'), Grid2OpException AmbiguousAction InvalidRedispatching InvalidRedispatching('Game over due to infeasible redispatching state. The routine used to compute the "next state" has diverged. This means that there is no way to compute a physically valid generator state (one that meets all pmin / pmax - ramp min / ramp max with the information provided. As one of the physical constraints would be violated, this means that a generator would be damaged in real life. This is a game over.')], 'rewards': {'_scores_grid_operational_cost': 1.0, '_scores_assistant_confidence': 0.0, '_scores_new_renewable_sources_usage': 1.0}}

BDonnot:
Ok so to avoid further issue like this, in the next PR (I'm waiting for the tests) I renamed the "old" tests, the one that passes and that should NOT be modified:
- `test_alert_obs_act` (previously `test_alert_feature`)
- `test_AlertReward` (previously `test_alert_score`)
- `test_score_idf_2023_nres` (previously `test_score_idf_2023`)

And the class `AlertReward` should not be modified either.

Unless a bug is spotted or a test there is missing, these files should not be modified by the current PR.

grid2op/tests/test_score_idf_2023.py
BDonnot: To avoid regression (which will happen if we continue to modify the tests like that to "make them pass") I renamed the test files:

- `test_alert_obs_act` (previously `test_alert_feature`)
- `test_AlertReward` (previously `test_alert_score`)
- `test_score_idf_2023_nres` (previously `test_score_idf_2023`)

And the class `AlertReward` should not be modified either.


Neither of these 4 files should be modified in this PR.

grid2op/utils/l2rpn_idf_2023_scores.py
BDonnot: The base class has constructor:

```python

    def __init__(
        self,
        env,
        env_seeds=None,
        agent_seeds=None,
        nb_scenario=16,
        min_losses_ratio=0.8,
        verbose=0,
        max_step=-1,
        nb_process_stats=1,
        scores_func=L2RPNSandBoxScore,
        score_names=None,
        add_nb_highres_sim=False,
    ):
   ```
   
   To avoid possible issues later on, in the derived class you should keep the same order. 

(also to avoid making print screen then copy paste the image...  you can use:

\`\`\`python 
python code here
 \`\`\`

It's easier and more efficient and other people can copy paste)

File: grid2op/utils/l2rpn_idf_2023_scores.py (hunk 16 above)
Comment (marota) [reply grid2op/utils/l2rpn_idf_2023_scores.py:123]: The weight_confidence_assistant_score is not used anymore (since the AlertCostScore is removed) so I delete it as argument and deleted this assert as well

File: grid2op/utils/l2rpn_idf_2023_scores.py (hunk 14 above)
Comment (marota) [reply grid2op/utils/l2rpn_idf_2023_scores.py]: Done

File: grid2op/tests/test_alert_trust_score.py [hunk 34]
```diff
@@ -0,0 +1,1753 @@
...
+
+                    total_nb_attacks = env._reward_helper.template_reward.total_nb_attacks
+                    nb_last_attacks = env._reward_helper.template_reward.nb_last_attacks
+
+                    assert nb_last_attacks == 0
+                    assert total_nb_attacks == 2
+
+                    cm_reward=env._reward_helper.template_reward.cumulated_reward
+                    assert env._reward_helper.template_reward.cumulated_reward==DEFAULT_PARAMS_TRUSTSCORE["reward_end_episode_bonus"] +\
+                            DEFAULT_PARAMS_TRUSTSCORE["reward_min_no_blackout"]*total_nb_attacks
```
laure-crochepierre, line 610: this assert should be : `env._reward_helper.template_reward.cumulated_reward==DEFAULT_PARAMS_TRUSTSCORE["reward_end_episode_bonus"] + DEFAULT_PARAMS_TRUSTSCORE["reward_min_no_blackout"]`

grid2op/Reward/_alertTrustScore.py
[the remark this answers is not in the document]
marota: I updated the behavior like this btw

        if(cm_reward_min_ep==cm_reward_max_ep):
            if (is_blackout):
                score_ep = 0.0
            else:
                score_ep = max_score
        else:
            score_ep = min_score + (max_score - min_score) * standardized_score

grid2op/Reward/alertReward.py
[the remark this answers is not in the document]
marota: Ok we get back to previous formulation as discussed

grid2op/tests/test_RewardAlertCostScore.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def attack(self, observation, agent_action, env_action, budget, previous_fails): 
+        if observation is None:
+            return None, None
+        current_step = self.env.nb_time_step
+        if current_step not in self.steps_attack: 
+            return None, None
+        
+        return self.custom_attack, self.duration
+
```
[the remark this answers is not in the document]
marota: Thanks for doing it

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1767 @@
...
+ATTACKED_LINE = "48_50_136"
+
+
+#DEFAULT_PARAMS_TRUSTSCORE = dict(reward_min_no_blackout=-1.0,
+#                                 reward_min_blackout=-10.0,
+#                                 reward_max_no_blackout=1.0,
+#                                 reward_max_blackout=2.0,
+#                                 reward_end_episode_bonus=42.0,
+#                                 min_score=-1.0)
+
```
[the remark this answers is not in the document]
marota: I removed it for now

grid2op/tests/test_alert_trust_score.py
marota: thanks for doing it

grid2op/tests/test_alert_trust_score.py
marota: thanks for doing it

grid2op/tests/test_alert_trust_score.py
marota: Yes no_blackout_2_attack_same_time tests are updated in the next commit I will do. This relates to how we decide to count the number of attacks which required a change in _alertTrustScore as well

grid2op/tests/test_alert_trust_score.py
marota: there is no l2rpn_idf_2023 in "data_test". Isn't it from where you load grid2op environments when doing tests ?
The l2rpn_idf_2023 in "data" has long scenarios. In case we use that one we should probably reduce the max_iter to the number of steps run in l2rpn_idf_2023_with_alert

BDonnot:
Can you: make the tests pass and clean the conversation that we have by marking "resolved" when the remark was adressed. 
I don't have the time to look again deeply into this :-/

grid2op/utils/l2rpn_idf_2023_scores.py
marota: Perhaps a good idea, keeping as it was set originally for now (probably Clement naming, should see with him)

grid2op/utils/l2rpn_idf_2023_scores.py
marota: Perhaps a good idea, keeping as it was set originally for now (probably Clement naming, should see with him)

grid2op/tests/test_score_idf_2023.py
marota: This was done in test_score_idf_2023_assistant.py

grid2op/tests/test_alert_trust_score.py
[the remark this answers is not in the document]
marota: The alertTrustScore with the runner is tested in test_score_idf_2023_assistant.py

grid2op/tests/test_alert_trust_score.py
marota: I suggest we keep the l2rpn_idf_2023_with_alert env for now in these tests as originally done in these tests when I took over and as far of knowledge I don't see better option so far. Could make a change of env later if relevant

grid2op/tests/test_score_idf_2023.py
marota: Ok I do so.
See that I need to comment data_feeding_kwargs as I get this error with l2rpn_idf_2023

Grid2op.Exceptions.handlers_exceptions.HandlerError: Grid2OpException ChronicsError HandlerError "No files are found in directory "/home/marotant/dev/Grid2Op_c400c90f/grid2op/data/l2rpn_idf_2023/chronics/2035-01-15_0/prod_vNone". If you don't want to load any chronics, use  "DoNothingHandler" (`from grid2op.Chronics.handlers import DoNothingHandler`) and not "<class 'grid2op.Chronics.handlers.csvHandler.CSVHandler'>" to load chronics."

grid2op/Reward/alertReward.py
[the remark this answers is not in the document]
marota: I do not touch this file so I guess it is done

marota:
Requested changes addressed, hopefully tests should pass

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,165 @@
...
+            #    self.no_alert_attack_no_blackout += np.sum(~lines_alerted_beforeattack[lines_attacked_no_blackout])
+            
+            return score_ep
+        
+    @staticmethod
+    def _normalisation_fun(cm_reward, cm_reward_min_ep, cm_reward_max_ep,min_score,max_score,is_blackout,tol=1e-5):
+        standardized_score = np.round((cm_reward - cm_reward_min_ep) / (cm_reward_max_ep - cm_reward_min_ep +tol),4)
+        #in case cm_reward_min_ep=cm_reward_max_ep=0, score is maximum
+        if(cm_reward_min_ep==cm_reward_max_ep):
+            if (is_blackout):
```
BDonnot, line 140: no parenthesis around "if" in python (and also add spaces around `==` )

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,165 @@
...
+            if(nb_last_attacks==0):
+                cm_reward_min_ep = self.reward_min_no_blackout * nb_attacks
+                cm_reward_max_ep = self.reward_max_no_blackout * nb_attacks
+            elif(nb_last_attacks>=1):
+                cm_reward_min_ep = self.reward_min_no_blackout * (nb_attacks - nb_last_attacks) + self.reward_min_blackout
+                cm_reward_max_ep = self.reward_max_no_blackout * (nb_attacks - nb_last_attacks) + self.reward_max_blackout
+        else:
+            cm_reward_min_ep = self.reward_min_no_blackout * nb_attacks
+            cm_reward_max_ep = self.reward_max_no_blackout * nb_attacks + self.reward_end_episode_bonus
+
```
BDonnot, line 163: don't understand all of that but i trust you. This looks really complicated with all these "ifs"

grid2op/tests/test_RewardAlertCostScore.py
```diff
@@ -0,0 +1,373 @@
...
+    
+    def setUp(self) -> None:
+        self.env_nm = os.path.join(
+            PATH_DATA_TEST, "l2rpn_idf_2023_with_alert"
+        )
+        
+    def tearDown(self) -> None:
+        return super().tearDown()
+        
+    def test_assistant_reward_value_no_blackout_no_attack_no_alert(self) -> None :
```
BDonnot, line 53: Only one test for this particular (and if I understood correctly really important) function ? this is ... very little

grid2op/tests/test_RewardAlertCostScore.py
```diff
@@ -0,0 +1,373 @@
...
+        with tempfile.TemporaryDirectory() as f:
+            res = runner.run(nb_episode=1, episode_id=[0], max_iter=10, env_seeds=[0],
+                             path_save=f)
+            assert res[0][2] == 1.
+            ep0, *_ = EpisodeData.list_episode(f)
+            ep = EpisodeData.from_disk(*ep0)
+            assert ep.rewards[8] == 1.
+ 
+
+class TestAlertTrustScore(unittest.TestCase):
```
BDonnot, line 174: I already said to make a test for Cost (in a separate file) and a test for Trust, in this file for example. It's really unclear what is done and where

grid2op/tests/test_RewardAlertCostScore.py
```diff
@@ -0,0 +1,373 @@
...
+                step += 1
+                
+                if step in _get_steps_attack(kwargs_opponent): 
+                    assert info["opponent_attack_line"] is not None, f"no attack is detected at step {step}"
+                else:
+                    assert info["opponent_attack_line"]  is None, f"an attack is detected at step {step}"
+                
+                if step == 4: 
+                    # When the blackout occurs, reward is -1 because we didn't raise an attack so min
+                    assert reward == env._reward_helper.template_reward.min_score, f"error for step {step}: {reward} vs -1"
```
BDonnot, line 259: don't rely on data that you want to test. Write clearly what is it supposed to be (-1 if I followed correctly here)

BDonnot (changes requested):
Too many changed files, can you restrict to 5 or 6 please.

grid2op/Reward/_alertCostScore.py
```diff
@@ -0,0 +1,75 @@
...
+    the assistant have produced during an episode. It should not be used to train an agent.
+    For information, it will not be used for the L2RPN_IDF_2023 competition.
+    """
+    def __init__(self, logger=None):
+        BaseReward.__init__(self, logger=logger)
+        self.reward_min = dt_float(-1.0)
+        self.reward_max = dt_float(1.0)
+        self._is_simul_env = False
+        self.total_nb_alertes_possible = None
+        self.total_nb_alerts = None
```
BDonnot, line 38: You can add this:

```python
warnings.warn("This class is not tested, use with care")
```

grid2op/tests/test_RewardAlertCostScore.py
```diff
@@ -0,0 +1,373 @@
...
+            res = runner.run(nb_episode=1, episode_id=[0], max_iter=10, env_seeds=[0],
+                             path_save=f)
+            assert res[0][2] == 1.
+            ep0, *_ = EpisodeData.list_episode(f)
+            ep = EpisodeData.from_disk(*ep0)
+            assert ep.rewards[8] == 1.
+ 
+
+class TestAlertTrustScore(unittest.TestCase):  
+    def setUp(self) -> None:
```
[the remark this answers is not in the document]
marota: see test_alert_trust_score.py

grid2op/tests/test_RewardAlertCostScore.py
marota: test_RewardAlertCostScore.py is disregarded for now

grid2op/tests/test_RewardAlertCostScore.py
marota: I removed everything from this file for AlertTrustScore, I reintegrated any non-overlapping tests in test_alert_trust_score.py

grid2op/tests/test_RewardAlertCostScore.py
marota: see test_alert_trust_score.py

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            score_ep = min_score + (max_score - min_score) * standardized_score
+        return score_ep
+    
+    def _compute_min_max_reward(self, nb_attacks,nb_last_attacks):
+
+        if (nb_attacks==0 and self.blackout_encountered):
+            cm_reward_min_ep = 0.
+            cm_reward_max_ep= 0.
+        elif(self.blackout_encountered):
```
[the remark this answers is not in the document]
marota: OK

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,165 @@
...
+            #    lines_attacked_no_blackout = env._time_since_last_attack > 0
+            #
+            #    self.alert_attack_no_blackout += np.sum(lines_alerted_beforeattack[lines_attacked_no_blackout])
+            #    self.no_alert_attack_no_blackout += np.sum(~lines_alerted_beforeattack[lines_attacked_no_blackout])
+            
+            return score_ep
+        
+    @staticmethod
+    def _normalisation_fun(cm_reward, cm_reward_min_ep, cm_reward_max_ep,min_score,max_score,is_blackout,tol=1e-5):
+        standardized_score = np.round((cm_reward - cm_reward_min_ep) / (cm_reward_max_ep - cm_reward_min_ep +tol),4)
```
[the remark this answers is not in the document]
marota: Previous answer "Because then you get score of 0.99993.
Where should we round this then ? Here it is to compensate for the addition of the tolerance that is artificial"

New answer: I remove the tol argument and rounding. cm_reward_max_ep == cm_reward_min_ep is a case that we deal with before, and don't call this _normalisation_fun if that's the case

marota:
> Too many changed files, can you restrict to 5 or 6 please.

Let me give you more insights on the list:

- grid2op/Reward/__init__.py: renaming of  "_AssistantConfidenceScore", "_AssistantCostScore" in  "_AlertCostScore",
    "_AlertTrustScore". This choice was made before I took over but nothing to worry about
    
- <del>alertCostScore.py<\del>: as discussed you can discard if for now as now used and I added warnings you suggested

- **alertTrustScore.py:** very important file for the assistant score to review, and you already did thanks

- _newRenewableSourcesUsageScore: this one I cannot trace back why this change is there (I haven't worked on it myself). I can only tell the only "change" is that it precomputes a max_timesteps variable that used twice after with the trade-off of one more line of code. This makes it a bit more readable. But we can definitely remove this change from this PR if you prefer. **What do you prefer ?**

- <del>test_RewardAlertCostSCore<\del>: as discussed you can discard if for now as now used and I added warnings you suggested. I also remove the alertTrustScore related tests and reintegrated non-overlapping ones in test_alert_trust_score

- test_RewardNewRenewableSourcesUsageScore.py: as for the related reward,  I cannot trace back why this change is there (I haven't worked on it myself). I can only tell it adds two more tests. This makes is more tested. But we can definitely remove this change from this PR if you prefer. **What do you prefer ?**

- test_alert_obs_act.py: it reintegrates a change that was forgotten as my last commit of your PR https://github.com/marota/Grid2Op/pull/5 explains

- **test_alert_trust_score.py**: the most extensive and very important test file to review that tests for the AlertTrustScore.

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,163 @@
...
+            # TODO
+            #lines_attacked_no_blackout = env._time_since_last_attack == SURVIVOR_TIMESTEPS
+#
+            #self.alert_attack_no_blackout += np.sum(lines_alerted_beforeattack[lines_attacked_no_blackout])
+            #self.no_alert_attack_no_blackout += np.sum(~lines_alerted_beforeattack[lines_attacked_no_blackout])
+            
+            return score_ep
+            
+        else:
+            self.nb_last_attacks=np.sum(np.any(self._ts_attack,axis=1))
```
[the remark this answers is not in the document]
BDonnot: Ok after a deep look it appears to be working because of the implementation of AlertReward which put to "False" the _ts_attack right after it has been processed by this class.

So maybe the bug is not where I thought it would be but I wonder now if self.nb_last_attacks can be 1  sometimes.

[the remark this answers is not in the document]
BDonnot: And it does not "break" because _ts_attack is NOT reset to `False`  in the base class when there is a game over.

So it's a real miracle that heavily depends on the behaviour of the implementation of the base class. This is not recommended.

At least use:

```python
            indexes_to_look = (np.arange(-self.time_window, 1) + self._current_id) % self._nrows_array  # include current step (hence the np.arange(..., **1**))
            ts_attack_in_order = self._ts_attack[indexes_to_look, :]
            self.nb_last_attacks = ts_attack_in_order.any(axis=1).sum()
```

So that it at least does not depend on the implementation of the base class

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,163 @@
...
+            score_ep = min_score + (max_score - min_score) * standardized_score
+        return score_ep
+    
+    def _compute_min_max_reward(self, nb_attacks,nb_last_attacks):
+
+        if self.blackout_encountered:
+            if nb_last_attacks == 0: #in the case the blackout is not tied to a recent attack
+                cm_reward_min_ep = self.reward_min_no_blackout * nb_attacks
+                cm_reward_max_ep = self.reward_max_no_blackout * nb_attacks
+            elif nb_last_attacks >= 1:#in the case one or several recent attacks can be tied to the blackout
```
BDonnot, line 155: blackout is caused by the attack, not the opposite (rephrase if possible)

grid2op/Reward/_alertTrustScore.py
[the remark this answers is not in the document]
BDonnot: A even better version would include a function (in AlertReward) that does the computation of `ts_attack_in_order` so that nothing is copy pasted.

grid2op/Reward/_alertTrustScore.py
BDonnot: `#in the case the blackout  can be tied to one or several recent attacks`

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1737 @@
...
+def manual_score (cm_reward,cm_reward_min_ep,cm_reward_max_ep,max_score):
+
+    manual_standardized_score= np.round((cm_reward - cm_reward_min_ep) / (cm_reward_max_ep - cm_reward_min_ep + 1e-5), 4)
+    manual_score = DEFAULT_PARAMS_TRUSTSCORE["min_score"] + (
+                                max_score - DEFAULT_PARAMS_TRUSTSCORE[
+                            "min_score"]) * manual_standardized_score
+    return manual_score
+
+# Test alertTrustScore when no blackout and when blackout
+class TestAlertTrustScoreNoBlackout(unittest.TestCase):
```
BDonnot, line 51: It's a a good start to have copy pasted all the tests in AlertReward and made them pass for this class. But it's not enough as i'm not sure it tests the new behaviour.

Especially, i don't thing you tried to find "corner cases" where the behaviour of the new class and the old one would differ. The previous tests were made for "corner cases" in the AlertReward. I'm not sure if there are the same "corner cases" for this class.

BDonnot: (though it's good to make sure there are no regressions on the code anywhere)

BDonnot: For example, the whole things is based on the  assumption that "whatever the case, the boundaries in `_normalisation_fun` can be achieved"

So it would be nice to include (or at least put together) tests that, for some "basic" situations (*eg* `blackout "due to" attack`, `blackout not due to attack`, `no blackout`) that you can design 3 agents : one achieving min score, one achieving max score and one performing "in the middle"

grid2op/tests/test_score_idf_2023_assistant.py
```diff
@@ -0,0 +1,166 @@
...
+                                    #                      "gen_p_handler": CSVHandler("prod_p"),
+                                    #                      "load_p_handler": CSVHandler("load_p"),
+                                    #                      "gen_v_handler": CSVHandler("prod_v"),
+                                    #                      "load_q_handler": CSVHandler("load_q"),
+                                    #                      "h_forecast": (5,),
+                                    #                      "gen_p_for_handler": PerfectForecastHandler("prod_p_forecasted", quiet_warnings=True),
+                                    #                      "gen_v_for_handler": PerfectForecastHandler("prod_v_forecasted", quiet_warnings=True),
+                                    #                      "load_p_for_handler": PerfectForecastHandler("load_p_forecasted", quiet_warnings=True),
+                                    #                      "load_q_for_handler": PerfectForecastHandler("load_q_forecasted", quiet_warnings=True),
+                                    #                    },
```
BDonnot, line 75: delete if not used

grid2op/utils/l2rpn_idf_2023_scores.py
BDonnot, line 10: remove file not needed

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -101,34 +109,33 @@ def __init__(
...
-        assert(weight_op_score + weight_assistant_score + weight_nres_score==1.)
-        assert(all([weight_confidence_assistant_score>=0., weight_confidence_assistant_score<=1.]))
-        
+
+        test_weights = weight_op_score + weight_assistant_score + weight_nres_score
+        if test_weights != 1.0:
+            raise Grid2OpException(
+                'The weights of each component of the score shall sum to 1'
+            )
+
```
BDonnot, line 122: and we don't test that everything weights is positive ?

BDonnot (changes requested):
Some minor modifications here and there.

Tests need to be reorganized I think. It's not clear (for me) that every "corner cases" for the score are met.
In particular, if the formula to compute min / max score (on which everything is based) is wrong, then the scores will be wrong. And I did not see explicit tests for it.

So what I suggest is that you put (with all the tests there, that are relevant as they ensure no regression compared to the AlertReward) 3 test classes:
- one where you will test blackout "caused by" attack
- one where you will test blackout "not caused by" attacks
- one where you will test "no blackout"

And in each of these classes, you make at least 3 tests:
- one where the agent reaches score min
- one where it reaches score max
- one where it reaches a score in the middle
- (of course other tests for other twisted situations can be a good idea)

Oh and the computation of `nb_last_attacks` in _AlertTrustScore really depends on some miracles. I suggested an implementation that does not rely that much on the actual implementation of AlertReward and the way the "_ts_attack" is reset and when it is 
/ when it is not.

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,163 @@
...
+        #self.no_alert_attack_blackout = 0
+        #self.blackout_encountered = False
+        return super().initialize(env)
+            
+    def reset(self, env):
+        super().reset(env)
+        self.cumulated_reward = 0
+        #KPIs
+        self.total_nb_attacks = 0
+        self.nb_last_attacks = 0
```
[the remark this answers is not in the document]
marota: Done

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,163 @@
...
+        #self.no_alert_attack_blackout = 0
+        #self.blackout_encountered = False
+            
+    def __call__(self, action, env, has_error, is_done, is_illegal, is_ambiguous):
+
+        score_ep = 0.
+        if self._is_simul_env:
+            return score_ep
+
+        self.blackout_encountered = self.is_in_blackout(has_error, is_done)
```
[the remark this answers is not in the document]
marota: Actually  this was duplicated once... 
 if self._is_simul_env:
            return score_ep
 So removing the second one, and  self.blackout_encountered is indeed after the mention of self._is_simul_env now

grid2op/Reward/_alertTrustScore.py
```diff
@@ -0,0 +1,163 @@
...
+        self.cumulated_reward += res
+
+        is_attack = (env._time_since_last_attack == 0).any()#even if there are simultaneous attacks, we consider this as a single attack event
+        self.total_nb_attacks += is_attack
+
+        # TODO
+        #lines_alerted_beforeattack = np.equal(env._time_since_last_alert, env._time_since_last_attack + 1) and lines_attacked
+        #self.total_nb_alerts += np.sum(lines_alerted_beforeattack)
+            
+        if not is_done:
```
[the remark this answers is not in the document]
marota: ok

grid2op/Reward/_alertTrustScore.py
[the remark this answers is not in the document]
marota: ok I added the code block for now

grid2op/Reward/_alertTrustScore.py
marota: Ok I did that

grid2op/tests/test_score_idf_2023_assistant.py
```diff
@@ -0,0 +1,166 @@
...
+            blackout_action.gen_set_bus = [(0, -1)]
+            act+=blackout_action
+
+        return act
+
+
+def get_blackout(self, env):
+    blackout_action = env.action_space({})
+    blackout_action.gen_set_bus = [(0, -1)]
+    return blackout_action
```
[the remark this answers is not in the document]
marota: ok

grid2op/tests/test_score_idf_2023_assistant.py
marota: ok

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+                    assert info["opponent_attack_line"] is not None, f"no attack is detected at step {step}"
+                else:
+                    assert info["opponent_attack_line"] is None, f"an attack is detected at step {step}"
+
+                if done:
+                    assert score == DEFAULT_PARAMS_TRUSTSCORE["min_score"]
+                    total_nb_attacks = env._reward_helper.template_reward.total_nb_attacks
+                    nb_last_attacks = env._reward_helper.template_reward.nb_last_attacks
+
+                    assert nb_last_attacks == 0
```
BDonnot, line 953: small comment maybe to say `  # because no attack caused the blackout`

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+        )
+
+    
+    def get_dn(self, env):
+        return env.action_space({})
+
+    def get_blackout(self, env):
+        blackout_action = env.action_space({})
+        blackout_action.gen_set_bus = [(0, -1)]
+        return blackout_action
```
BDonnot, line 1172: same as above, avoid copy paste and prefer reusing the _get_blackout function defined in the _aux_opponent_for_test_blablabla

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+                    assert info["opponent_attack_line"] is not None, f"no attack is detected at step {step}"
+                else:
+                    assert info["opponent_attack_line"]  is None, f"an attack is detected at step {step}"
+                
+                if done:
+                    assert score == env._reward_helper.template_reward.max_score
+                    total_nb_attacks = env._reward_helper.template_reward.total_nb_attacks
+                    nb_last_attacks = env._reward_helper.template_reward.nb_last_attacks
+
+                    assert nb_last_attacks == 1
```
BDonnot, line 1225: small comment here to say `  # because blackout caused by attack`

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+                    act = env.action_space({"raise_alert": [attackable_line_id]})
+                obs, score, done, info = env.step(act)
+                step += 1
+                if step in _get_steps_attack(kwargs_opponent): 
+                    assert info["opponent_attack_line"] is not None, f"no attack is detected at step {step}"
+                else:
+                    assert info["opponent_attack_line"]  is None, f"an attack is detected at step {step}"
+                
+                if done: 
+                    assert score == DEFAULT_PARAMS_TRUSTSCORE["min_score"]
```
BDonnot, line 1286: small comment here to say   `# because blackout caused by attack`

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+                if step in _get_steps_attack(kwargs_opponent):
+                    assert info["opponent_attack_line"] is not None, f"no attack is detected at step {step}"
+                else:
+                    assert info["opponent_attack_line"] is None, f"an attack is detected at step {step}"
+
+                if done:
+                    total_nb_attacks = env._reward_helper.template_reward.total_nb_attacks
+                    nb_last_attacks = env._reward_helper.template_reward.nb_last_attacks
+
+                    assert nb_last_attacks == 1
```
BDonnot, line 1355: small comment here to say   `# because blackout caused by attack`

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+                elif i == 1:
+                    act = env.action_space({"raise_alert": [0]})
+                obs, score, done, info = env.step(act)
+                if info["opponent_attack_line"] is None:
+                    if done:  # info["opponent_attack_line"] is None :
+                        assert score == 0.
+                        total_nb_attacks = env._reward_helper.template_reward.total_nb_attacks
+                        nb_last_attacks = env._reward_helper.template_reward.nb_last_attacks
+
+                        assert nb_last_attacks == 0
```
BDonnot, line 997: small comment maybe to say   `# because no attack caused the blackout`

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+                    assert info["opponent_attack_line"] is not None, f"no attack is detected at step {step}"
+                else:
+                    assert info["opponent_attack_line"] is None, f"an attack is detected at step {step}"
+
+                if done:
+                    assert score == env._reward_helper.template_reward.max_score
+                    total_nb_attacks = env._reward_helper.template_reward.total_nb_attacks
+                    nb_last_attacks = env._reward_helper.template_reward.nb_last_attacks
+
+                    assert nb_last_attacks == 0
```
BDonnot, line 887: small comment maybe to say   `# because no attack caused the blackout`

grid2op/utils/l2rpn_idf_2023_scores.py
```diff
@@ -101,34 +110,37 @@ def __init__(
...
-        assert(weight_op_score + weight_assistant_score + weight_nres_score==1.)
-        assert(all([weight_confidence_assistant_score>=0., weight_confidence_assistant_score<=1.]))
-        
+        weights=np.array([weight_op_score,weight_assistant_score,weight_nres_score])
+        test_weights = weights.sum()
+        if test_weights != 1.0:
+            raise Grid2OpException(
+                'The weights of each component of the score shall sum to 1'
+            )
+        if np.any(test_weights <0):
```
BDonnot, line 123: `np.any(weights)` (or even better `weitghts.any()`

BDonnot: (and not test_weights)

BDonnot (changes requested):
Almost there: some minor things to change in the tests, some comments to add there and a typo to fix in the l2rpn_idf_2023_scores.py

grid2op/utils/l2rpn_idf_2023_scores.py
marota: Done

grid2op/tests/test_alert_trust_score.py
```diff
@@ -0,0 +1,1892 @@
...
+        )
+
+    def get_dn(self, env):
+        return env.action_space({})
+
+    def get_blackout(self, env):
+        blackout_action = env.action_space({})
+        blackout_action.gen_set_bus = [(0, -1)]
+        return blackout_action
+
```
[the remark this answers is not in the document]
marota: Ok option B) taken

BDonnot (approved):
Looks good, thanks @GoubetClem  @laure-crochepierre  and @marota

Issue #522, opened 2023-09-14
Observation notebook does not work


benedikt-schesch:
## Environment
 - Grid2op version: `a819a777cc2da67270c84f1887f8cb32fec23308`
 - System: `osx`

## Bug description

get started guide 02_Observation does not execute properly due to missing proper default value.

## How to reproduce
Simply run `getting_started/02_Observation.ipynb` you will see that block 4 fails with the following message:

```
---------------------------------------------------------------------------
Grid2OpException                          Traceback (most recent call last)
Cell In[4], line 1
----> 1 env = grid2op.make(test=True)
      2 obs = env.reset() 

File [~/Git/Grid2Op/getting_started/grid2op/MakeEnv/Make.py:350](https://file+.vscode-resource.vscode-cdn.net/Users/benediktschesch/Git/Grid2Op/getting_started/~/Git/Grid2Op/getting_started/grid2op/MakeEnv/Make.py:350), in make(dataset, test, logger, experimental_read_from_local_dir, _add_to_name, _compat_glop_version, **kwargs)
    347         test = True
    349 if dataset is None:
--> 350     raise Grid2OpException("Impossible to create an environment without its name. Please call something like: \n"
    351                            "> env = grid2op.make('l2rpn_case14_sandbox') \nor\n"
    352                            "> env = grid2op.make('rte_case14_realistic')")
    354 accepted_kwargs = ERR_MSG_KWARGS.keys() | {"dataset", "test"}
    355 for el in kwargs:

Grid2OpException: Grid2OpException "Impossible to create an environment without its name. Please call something like: 
> env = grid2op.make('l2rpn_case14_sandbox') 
or
> env = grid2op.make('rte_case14_realistic')"
```

Problem is that the default value is None which is not what is written in the notebook

benedikt-schesch:
I would very highly recommend adding the execution of all the notebooks to the CI/CD pipeline so changes to the API don't break the user guides

benedikt-schesch:
This problem is solved by #524

BDonnot:
> I would very highly recommend adding the execution of all the notebooks to the CI/CD pipeline so changes to the API don't break the user guides

Hello,

You are correct here, this is what used to be done but at some point I had trouble with their execution (though the notebooks worked perfectly, the CI still send weird error messages) and this is why I deactivated it. There is an issue about that #234 that is not on my priority at the moment (and I need to redo some of the notebooks...)
Too many things to do, not enough time to do it apparently...

[...]

PR #524, opened 2023-09-14
Added proper default value to 'make' that create the envs

benedikt-schesch:
Adds a proper default value to the creation of the environment. Also makes the type hints correct.


benedikt-schesch:
Resolves issue #522

BDonnot:
Hello,

I disabled the default environment (for clarity) in previous grid2op release : I don't think it's a good idea to have something by default, especially this environment (`l2rpn_case14_sandbox` would be a much better but lacking lots of features)

I think the best approach would be to force the first argument, for example with removing the `=None` (this is definitely what I should have done)

Thanks for this PR but as is I am not sure i will merge it (though i'm open for discussion on the default argument to the dataset_path of course and I can change my mind about it)

benedikt-schesch:
@BDonnot I think you are right, there should be no default value. I changed the typing and the notebook to reflect this correctly.

BDonnot:
can you make it to dev_1.9.6 now please ? I made a realease to fix some bugs (including the high priority one you spotted) yesterday.
Thanks :-)

benedikt-schesch:
@BDonnot There you go

Issue #527, opened 2023-09-15
Action space sampling returns invalid actions


nicow-elia:
## Environment
 - Grid2op version: `1.9.3`
 - System: `ubuntu16.04`
 - Additional system information

## Bug description
When sampling from the action space and converting these actions into serializable dicts so they can be loaded from disc later, we will occasionally get results from the sort `{"change_bus": {}}` which is a do-nothing action and should be `{}`. Since grid2op 1.9, this type of action is not supported as input to the environment and raises an exception

## How to reproduce
```
import grid2op

def test_action_space_sampling() -> None:
    env = grid2op.make("l2rpn_case14_sandbox")
    for _ in range(10000):
        act_dict = env.action_space.sample().as_serializable_dict()
        env.action_space(act_dict)
```

## Current output

```
The test raises an exception
```

## Expected output

```
The test succeeds.
```

BDonnot:
Hello,

Thanks for spotting this. I fixed the issue in two different ways:

1) the as_serializable_dict() method will now discard the keys associated with "do nothing" (when nothing is modified for example)
2) the action will be nicer and accept "weird" do nothing actions such as `env.action_space({"change_bus": {}})`

I expect to release a new grid2op version next Monday (it's always a bad idea to release anything at all on Fridays)

PR #528, opened 2023-09-15
fix issue #527

BDonnot:


BDonnot (commented):
looks good (if tests pass)

PR #529, opened 2023-09-18
first working POC on ieee grid and for backend actions, no regular action, no observation yet

BDonnot:


EloyAnguiano:
Does this feature lead to be able to operate each substation by its switches? It would be interesting to be able to make an agent that deals with the interruptors of each substation independently

BDonnot:
At term it would yes. But for now it's far for complete and I can't work it at the moment (I thought it would be necessary for the backend your are working on but apparently it's not).

Also I would argue against using switches for RL. There are much more switches with much more intricate and non linear impact than the current modelling of the topology.

Issue #538, opened 2023-10-10
EpisodeData to/from disk does not preserve energy graphs


DEUCE1957:
## Environment
 - Grid2op version: `1.9.5`
 - System: Windows 11
 - Python 3.11.3

## Bug description
When loading a previously completed Episode from_disk using GridOp's EpisodeData class, the observations return an empty energy graph (1 node, no edges). However if we plot the Observation's EnergyGraph before saving to disk (i.e. within the RL loop) it returns a valid energy graph. It seems obs.get_energy_graph() relies on information that is not stored in obs.to_vect() (specifically obs.flow_bus_matrix()).

## How to reproduce
The following code runs a minimum EpisodeData class (without a runner) for 1 timestep on the "rte_case14_realistic" environment, we plot the Energy Graph for the observation before and after saving from disk.
```python
import logging
import grid2op
import numpy as np
import networkx as nx
import matplotlib.pyplot as plt
from grid2op.Episode import EpisodeData
from pathlib import Path as p
env = grid2op.make("rte_case14_realistic")
obs = env.reset()

max_no_of_timesteps = 1

# Make Temporary Path to Store Episode Data in
path_save = p.cwd() / "Temp"
path_save.mkdir(exist_ok=True)

array_names = ("actions", "env_actions", "observations", "attacks", "disc_lines", "rewards", "times")

# Arrays needed to initialize Episode Data class
episode_data_dict = dict( 
    path_save=path_save,
    logger = logging.Logger("Logger"),
    actions = np.full((max_no_of_timesteps, env.action_space.n), fill_value=np.NaN, dtype=np.float16),
    env_actions = np.full((max_no_of_timesteps, env._helper_action_env.n), fill_value=np.NaN, dtype=np.float32),
    observations = np.full((max_no_of_timesteps + 1, len(obs.to_vect())),fill_value=np.NaN,dtype=np.float32),
    rewards = np.full(max_no_of_timesteps, fill_value=np.NaN, dtype=np.float32),
    disc_lines = np.full((max_no_of_timesteps, env.backend.n_line), fill_value=np.NaN, dtype=np.bool_),
    times = np.full(max_no_of_timesteps, fill_value=np.NaN, dtype=np.float32),
    disc_lines_templ = np.full((1, env.backend.n_line), fill_value=False, dtype=np.bool_),
    attack_templ = np.full((1, env._oppSpace.action_space.size()), fill_value=0.0, dtype=np.float32),
[...]

Issue #429, opened 2023-03-17
Have a "class attribute" of the Backend to tell which file formats are handled.


BDonnot:
When people creates the backend, or the BackendClass they fill this and then grid2op checks for the right file at the right format

PR #547, opened 2023-10-26
Version 1.9.6

BDonnot:
[1.9.6] - 2023-10-26
----------------------
- [BREAKING] when a storage is connected alone on a bus, even if it produces / absorbs 0.0 MW it 
  will raise a diverging powerflow error (previously the storage was automatically disconnected by 
  `PandaPowerBackend`, but probably not by other backends)
- [BREAKING] when a shunt is alone on a bus, the powerflow will diverge even in DC mode 
  (previously it only converges which was wrong behaviour: grid2op should not disconnect shunt)
- [FIXED] a bug in PandaPowerBackend (DC mode) where isolated load did not raised 
  exception (they should lead to a divergence)
- [FIXED] some wrong behaviour in the `remove_line_status_from_topo` when no observation where provided
  and `check_cooldown` is `False`
- [FIXED] a bug in PandaPowerBackend in AC powerflow: disconnected storage unit had no 0. as voltage
- [FIXED] a bug in PandaPowerBackend in AC powerflow when a generator was alone a bus it made the powerflow
  crash on some cases (*eg* without lightsim2grid, without numba)
- [FIXED] a bug in PandaPowerBackend in DC (in some cases non connected grid were not spotted)
- [FIXED] now the observations once reloaded have the correct `_is_done` flag (`obs._is_done = False`)
  which allows to use the `obs.get_energy_graph()` for example. This fixes https://github.com/rte-france/Grid2Op/issues/538
- [ADDED] now depends on the `typing_extensions` package
- [ADDED] a complete test suite to help people develop new backend using "Test Driven Programming" 
  techniques
- [ADDED] the information on which time series data has been used by the environment in the `info`return value
  of `env.step(...)`
- [ADDED] a test suite easy to set up to test the backend API (and only the backend for now, integration tests with
  runner and environment will follow)
- [ADDED] an attribute of the backend to specify which file extension can be processed by it. Environment creation will
  fail if none are found. See `backend.supported_grid_format` see https://github.com/rte-france/Grid2Op/issues/429
- [IMPROVED] now easier than ever to run the grid2op test suite with a new backend (for relevant tests)
- [IMPROVED] type hints for `Backend` and `PandapowerBackend`
- [IMPROVED] distribute python 3.12 wheel
- [IMPROVED] test for python 3.12 and numpy 1.26 when appropriate (*eg* when numpy version is released)
- [IMPROVED] handling of environments without shunts
- [IMPROVED] error messages when grid is not consistent 
- [IMPROVED] add the default `l2rpn_case14_sandbox` environment in all part of the docs (substituing `rte_case14_realistic` or nothing)
- [IMPROVED] imports on the `Exceptions` module
- [IMPROVED] pandapower backend raises `BackendError` when "diverging"


BDonnot (commented):
Looks good, provided the changes in the next commit

PR #549, opened 2023-10-26
Allow simple customization of the action and observation space

benedikt-schesch:


benedikt-schesch:
@BDonnot Do you think this doesn't cause a problem on the lower layers? If you think it's good, I would also modify the docs to mention this before merging.

BDonnot:
Hello,

A few remarks remarks here (before a more in depth review)

Can you make your pr to dev_1.9.7 branch instead?

And also if tests do not pass then it cannot be merged. Do you have an idea on the failing tests?

grid2op/gym_compat/gym_act_space.py
```diff
@@ -115,11 +116,27 @@ class __AuxGymActionSpace:
...
     keys_human_2_grid2op = {v: k for k, v in keys_grid2op_2_human.items()}
 
-    def __init__(self, env, converter=None, dict_variables=None):
+
+    def __init__(
+        self,
+        env,
+        converter=None,
+        dict_variables=None,
+        action_attr_to_keep: Union[List[str],None] = None,
```
BDonnot, line 125: Variable name is `attr_to_keep` in box_gym_actsspace

grid2op/gym_compat/gym_obs_space.py
```diff
@@ -86,7 +87,12 @@ class __AuxGymObservationSpace:
 
     """
 
-    def __init__(self, env, dict_variables=None):
+    def __init__(
+        self,
+        env,
+        dict_variables=None,
+        obs_attr_to_keep: Union[List[str],None] = None,
```
BDonnot, line 94: Variable name is `attr_to_keep` in box_gym_obsspace

grid2op/gym_compat/gym_act_space.py
```diff
@@ -115,11 +116,27 @@ class __AuxGymActionSpace:
     }
     keys_human_2_grid2op = {v: k for k, v in keys_grid2op_2_human.items()}
 
-    def __init__(self, env, converter=None, dict_variables=None):
+
+    def __init__(
+        self,
+        env,
+        converter=None,
+        dict_variables=None,
```
BDonnot, line 124: Isn't this variable, named `dict_variables` doing what you want to do ? (if so it would be clearer to just rename it, no ?)

BDonnot: same for observation by the way :-)

BDonnot (commented):
Hello :-)

Thanks for this PR.

I have a few questions here:
1) isn't the `dict_variables` kwargs achieving what you want ? If so can you simply rename it maybe ?
2) tests are missing
3) doc is missing (for this new feature). I know the docs of the "gym_compat" module is far from perfect but it would still be better to have the doc for the new feature I think :-)

Thanks again

Benjamin

PR #575, opened 2024-01-17
fix normalization warning

EBoguslawski:
When using a gym action space or a gym observation space, a warning is raised if we try to normalize an attribute that can not be normalized with the default bounds. For exemple if the bounds are infinite or if they are the same. The warning printed a boolean vector instead of the attribute's name. It is now fixed.
There was also an error in the __AuxBoxGymActSpace class example.


BDonnot:
Looks good, thanks :-)

Issue #577, opened 2024-01-18
Wrong importation when importing a file named "utils.py" after grid2op


EBoguslawski:
## Environment
 - Grid2op version: the behavior appears between`1.9.6` and `1.9.8` and not before `1.9.5`
 - System: `windows` and `ubuntu16.04`

## Bug description

If you try to import from a file named `utils.py` after having imported `grid2op`, you don't import the functions in that `utils.py` file. You actually import the functions in `grid2op.utils`.

To get around the problem, a solution is of course to rename the `utils.py` file. We can also import `utils` before `grid2op`. This issue is not a priority, but it comprises some codes that were working with older versions of `grid2op`.

## How to reproduce

Create a parent folder with three python scripts like this :

```
parent_folder /
- utils.py
- utils_v2.py
- script.py
```

The file `utils.py` contains the code:

```
def print_hello_world():
    print("Hello World")
```

The file `utils_v2.py` contains the code:

```
def print_hello_world_v2():
    print("Hello World v2")
```

The file `script.py` contains the code:

```
import grid2op
from utils import *
from utils_v2 import *

print(dir(utils))
print(dir(utils_v2))
```

Run the `script.py` file.

## Current output with a version `>= 1.9.6`
```
['EpisodeStatistics', 'ScoreICAPS2021', 'ScoreL2RPN2020', 'ScoreL2RPN2022', 'ScoreL2RPN2023', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'print_hello_world_v2'] 
```

## Expected output

```
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'print_hello_world'] 
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'print_hello_world_v2'] 
```

BDonnot:
Solved in 1.9.8 released :-)

EBoguslawski:
It works for me now, thanks :)

Issue #571, opened 2023-12-15
Prevent the environment from disconnecting a chosen set of lines even in case of overflow


EBoguslawski:
## Description of the feature request

By default, the environment disconnects the powerlines above their thermal limit. **I need a feature that prevents the environment from disconnecting a chosen set of lines even if their flow is above their thermal limit.**

## Additional context

I need to train an agent that handles a specific zone of the power grid. The agent only cares about the safety of its zone. Overflows outside its zone do not concern it and would interfere with its training.
I need this feature to avoid blackouts caused by an overflow on the rest of the grid.

BDonnot:
Hello,

You can use the new feature by:

First nstalling grid2op from source from the github `https://github.com/rte-france/Grid2Op.git` (it should show version 1.9.8.dev0)

Then you can create a masked environment with:

```python
import grid2op
from grid2op.Environment import MaskedEnvironment

lines_I_care_about = np.zeros(n_line, dtype=bool)
lines_I_care_about[IDs of these lines] = True

env = MaskedEnvironment(grid2op.make(env_name),
                        lines_of_interest=lines_I_care_about))
```

I did not completely test it but it should be working relatively well, i'll try to test it more in depth ASAP

EBoguslawski:
Hello Benjamin,

Thank you a lot for making this feature available so quickly !

:)

BDonnot:
Added in 1.9.8 version :-)

Issue #569, opened 2023-12-08
Creating Gif via replay episode returns an error 


ML-IEE:
## Environment
 - Grid2op version: `1.9.5`

## Bug description
Hey, I am not quite sure whether or not this might already been fixed in the newer versions, however in Grid2Op v1.9.5 I was not able to create a gif due to the depriciated argument "fps". I could not see in the Release note, whether this was fixed. 

## How to reproduce
The code is quite similar to the [plotting example](https://grid2op.readthedocs.io/en/latest/plot.html)
In the following example we import the results from an Runner 
```python
import grid2op
from grid2op.Episode import EpisodeReplay
env = grid2op.make("l2rpn_wcci_2022")

plot_epi = EpisodeReplay("path")
plot_epi.replay_episode("2050-08-08_5",gif_name = "08-08-05",
                        start_step = 385, end_step = 700)
```

## Possible Error 
As far as I can see the error is due to the depreciated fps argument in the .replay_episode() method
```python
imageio.mimwrite(gif_path, frames, fps=fps)
```
## Proposed fix
The fix is rather simple. Just use duration instead of fps 
```python
imageio.mimwrite(gif_path, frames, duration=1000/fps)
```

BDonnot:
Hi, 

Thanks for noticing it. I added a `try / catch` block to make sure the old behaviour is still correct on most ancient imageio version, but beside that your fix seems good :-) 

Thanks !

BDonnot:
Should be solved in 1.9.8

PR #579, opened 2024-01-26
Dev 1.9.8

BDonnot:
- [FIXED] the `backend.check_kirchoff` function was not correct when some elements were disconnected 
  (the wrong columns of the p_bus and q_bus was set in case of disconnected elements)
- [FIXED] `PandapowerBackend`, when no slack was present
- [FIXED] the "BaseBackendTest" class did not correctly detect divergence in most cases (which lead 
  to weird bugs in failing tests)
- [FIXED] an issue with imageio having deprecated the `fps` kwargs (see https://github.com/rte-france/Grid2Op/issues/569)
- [FIXED] adding the "`loads_charac.csv`" in the package data
- [FIXED] a bug when using grid2op, not "utils.py" script could be used (see 
  https://github.com/rte-france/Grid2Op/issues/577). This was caused by the modification of
  `sys.path` when importing the grid2op test suite.
- [ADDED] A type of environment that does not perform the "emulation of the protections"
  for some part of the grid (`MaskedEnvironment`) see https://github.com/rte-france/Grid2Op/issues/571
- [ADDED] a "gym like" API for reset allowing to set the seed and the time serie id directly when calling
  `env.reset(seed=.., options={"time serie id": ...})`
- [IMPROVED] the CI speed: by not testing every possible numpy version but only most ancient and most recent
- [IMPROVED] Runner now test grid2op version 1.9.6 and 1.9.7
- [IMPROVED] refacto `gridobj_cls._clear_class_attribute` and `gridobj_cls._clear_grid_dependant_class_attributes`
- [IMPROVED] the bahviour of the generic class `MakeBackend` used for the test suite.
- [IMPROVED] re introducing python 12 testing
- [IMPROVED] error messages in the automatic test suite (`AAATestBackendAPI`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,27 +3,27 @@ version: 2.1
 executors:
     grid2op-executor:
         docker:
-            - image: python:3.10-buster
+            - image: python:3.10
         working_directory: /Grid2Op
 
     python37:
         docker:
-            - image: python:3.7-buster
+            - image: python:3.7
     python38:
         docker:
-            - image: python:3.8-buster
+            - image: python:3.8
     python39:
         docker:
-            - image: python:3.9-buster
+            - image: python:3.9
     python310:
         docker:
-            - image: python:3.10-buster
+            - image: python:3.10
     python311:
         docker:
-            - image: python:3.11-buster
+            - image: python:3.11
     python312:
         docker:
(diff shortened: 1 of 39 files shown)


BDonnot (commented):
looks good

PR #585, opened 2024-02-29
Mdp doc (init) and finish tests on more than 2 busbar per substation 

BDonnot:


BDonnot (commented):
looks good beside a few typos in the doc (will be changed in next PR)

PR #556, opened 2023-11-17
Add: Compact Episode Data

DEUCE1957:
Based on suggestion in Issue: https://github.com/rte-france/Grid2Op/issues/538
Adds an alternative way to store Episode Data, either through the Runner class or independently (see example in DocString). Instead of creating a folder for each episode, CompactEpisodeData creates a single compressed numpy archive that contains all the arrays. Other rewards are also stored as a 2d numpy array (instead of a list of dictionaries).

The interface for CompactEpisodeData is different from EpisodeData (we pass the environment and initial observation to it upon initialization, for instance) which means we don't need to create all the arrays inside _aux_run_one_episode(...). 

I duplicated and tweaked the EpisodeData tests to also run for CompactEpisodeData. Since the CompactEpisodeData does not use the CollectionWrapper, you cannot call .get_types() directly on the actions (but you can still grab the action_space and do it that way).

Other ideas for improving CompactEpisodeData (not implemented):
* Store Episode Metadata in 1 file per environment, rather than 1 file per episode
* Add documentation for how to use CompactEpisodeData
* Eventually merge CompactEpisodeData into EpisodeData (or vice versa), since they serve similar functions (also avoids needing a bunch of if statements and an extra keyword argument inside the Runner class)

PR #592, opened 2024-03-12
Add: Compact Episode Data (Debugged)

DEUCE1957:
Based on suggestion in Issue: https://github.com/rte-france/Grid2Op/issues/538
New pull request since old PR was for Grid2op 1.9.7 ( https://github.com/rte-france/Grid2Op/pull/556 )
Debug Note: When setting a single seed the env.seed_used attribute would return an array of form nd.array(0). Fixed by using .item() in this edge case.

Adds an alternative way to store Episode Data, either through the Runner class or independently (see example in DocString). Instead of creating a folder for each episode, CompactEpisodeData creates a single compressed numpy archive that contains all the arrays. Other rewards are also stored as a 2d numpy array (instead of a list of dictionaries).

The interface for CompactEpisodeData is different from EpisodeData (we pass the environment and initial observation to it upon initialization, for instance) which means we don't need to create all the arrays inside _aux_run_one_episode(...). 

I duplicated and tweaked the EpisodeData tests to also run for CompactEpisodeData. Since the CompactEpisodeData does not use the CollectionWrapper, you cannot call .get_types() directly on the actions (but you can still grab the action_space and do it that way).

Other ideas for improving CompactEpisodeData (not implemented):
* Store Episode Metadata in 1 file per environment, rather than 1 file per episode
* Add documentation for how to use CompactEpisodeData
* Eventually merge CompactEpisodeData into EpisodeData (or vice versa), since they serve similar functions (also avoids needing a bunch of if statements and an extra keyword argument inside the Runner class)

### DIFF
--- a/grid2op/Episode/CompactEpisodeData.py
+++ b/grid2op/Episode/CompactEpisodeData.py
@@ -1,0 +1,312 @@
+# Copyright (c) 2019-2020, RTE (https://www.rte-france.com)
+# Addition by Xavier Weiss (@DEUCE1957)
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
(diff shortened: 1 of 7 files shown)


BDonnot (approved):
looks good, thanks :)

PR #635, opened 2024-09-20
Dev switches

BDonnot:
Add detailed topology (alpha version)

Lots of optimization to do and some missing feature, especially in the "workflows":

- if an agent only uses switches and backend only want switch, do not compute the "nodal topology"
- conversly if the backend wants nodal topology and agent only uses "nodal type" action do not bother with switches
- you can cache the conversion nodal -> switches also
- need to do documentation
- need to perform more test
- need to include the observation now
- readme
- changelog

Lots of things to be done


BDonnot:
looks good

PR #670, opened 2024-11-29
Fix bug in the MultifolderWithCache.seed method

EBoguslawski:
If you use `MultifolderWithCache` to handle chronics and if you filter chronics to keep chronics with high indexes (indexes higher than the number of selected chronics), then you obtain an "index out of bounds" bug.
This PR adresses this bug by modifying the shape on the concerned array and adding a test.


BDonnot (approved):
Looks and thanks for the added tests

PR #687, opened 2025-03-05
[WIP] Create Protection and ThermalLimits classes to delegate calculations from the backend

YacineAbd:


BDonnot:
Hello,

Thanks for this.

Can you please :
- sign your commit (see some details on the dco bot)
- change the title to include "[WIP]" in it (to prevent accidental merge before all tests pass for example) 

Thanks 😊

BDonnot (changes requested):
We cannot merge it in the correct branch.

And there are some other tests that need to be made, especially for the soft overflow but also for some corner cases (see my comments above)

BDonnot:
This work might be picked-up later. I'm closing it in the mean time.

PR #708, opened 2025-05-09
Improve the "info" part of env.step

BDonnot:


BDonnot (commented):
Looks good even if codacy complains...

PR #725, opened 2025-09-18
Demand Response / Flexibility in Loads

DEUCE1957:
Attempt to extend Redispatching by enabling Environment's to have flexible loads. See "getting_started" / "13_Demand?Response.ipynb" for example use. Related to issue https://github.com/Grid2op/grid2op/issues/623

Currently working:
Able to define "flexibility" actions (env.action_space({"flexibility":[("load_0_0", 0.1]}) with correct print string for BaseAction
Able to load flexible load attributes from "flex_loads_charac.csv"
"InvalidFlexibility" error if flexibility exceeds size of load
Observable change in load after executing flexibility action in environment
Integration with redispatch routine (flexibility and redispatch are solved together in one optim. problem)
Observable change in actual/target flexibility after executing flexibility action in environment
Automated tests

Open Challenges/Issues:

    How should we define the max load / load size? (currently is max of all chronix2grid load profiles in all of the chronics)

Progress

    Actual/Target Flexibility in BaseObservation
    Flexibility in BaseAction / PlayableAction
    Flexibility in GridObject
    Flexibility in SerializableActionSpace.sample()

Conventions:
Positive flexibility: Reduce consumption (less demand in the grid)
Negative flexibility: Increase consumption (more demand in the grid)
target_flex: Sum of all flexibility at a load
actual_flex: Sum of all physically feasible flexibility at a load (determined by Environment)


DEUCE1957:
Hej Benjamin apologies for the late response (was at a conference).

```
Does having a ramping up of 0 would mean this is not possible to "reconnect" the loads to the grid ?

For example, if you "curtail" 10 MW of loads at 09:00 maybe the issue that lead to this measure disappeared. And you can "reconnect" these 10MW at say 13:00 for example. We definitely want to be able to do this.
```
Aye with the "positive flexibility" it is possible to 'undo' previous curtailments of load. So if I ask for -10MW at 09:00 (consume 10 MW less than I would have otherwise) I can ask for +10MW at 13:00 to restore it to the original load. 

```
However, the feature you target is really interesting. And I think it would be better to frame it as "ratio" (like curtailment) that would act both on load_p and load_q at the same time. And of course the ratio would be limited between 0 and 1 (or 0% and 100%).
```
I do like that idea! Perhaps like curtailment we can have both a mw and ratio version. I can reformulate the current as flexibility_mw, get it to pass the tests and then add flexibility (without '_mw') as the ratio? Will need to see if the logic needs to be moved from redispatch to curtailment then.

DEUCE1957:
ToDo:
- [ ]  Fix 'allowLoadFlexibility' not being appended when loading generated class files when using grid2op.make(..., class_in_file=True)

PR #736, opened 2026-02-04
Dev 1.12.3

BDonnot:
[1.12.3] - 2026-02-04
-----------------------
- [FIXED] the warnings when building the documentation.
- [FIXED] the deprecation warnings when importing grid2op 
  with recent python versions (due to presence of math equation 
  in some docstring)
- [FIXED] issues when loading a grid with disconnected elements: grid2op
  did not know on which bus to reconnect them when only the "reconnect" bus was given.
- [FIXED] an issue leading to wrong setpoint values for shunt_p and shunt_q in the previous
  stored state (EnvPreviousState)
- [ADDED] a test (in the AAA test) to assess that the backend._sh_vnkv is properly set if the shunts are
  handled by the backend.
- [IMPROVED] code for AAA backend tests (avoid equality check for float)
- [IMPROVED] doc when loading grid with disconnected elements


BDonnot (commented):
ok

Issue #738, opened 2026-02-12
Load environment with custom config.py file not in root directory


kostialjn:
## Is your feature request related to a problem? Please describe.

I would like to be able to point to a custom config.py file at grid2op.make
For the same grid I have several config.py files and I would like to be able to switch config without having to duplicate my environment folder

## Describe the solution you'd like

Ex. This is how i would like it to be done:
```python
import grid2op
env_name = "environment/folder/"
env_conf_1 = grid2op.make(env_name, _config_path="environment/folder/configs/config_1.py")
env_conf_2 = grid2op.make(env_name, _config_path="environment/folder/configs/config_2.py")
```

## Describe alternatives you've considered

Other alternatives : 
- Duplicate entire environment folder
- Import config files using `importlib` and load env using `grid2op.make(env_name, **config_data)`

Both those alternatives seem suboptimal

BDonnot:
Hello,

Today it seems to be relatively hard to do. Indeed, in multiple places in grid2op there are calls like "from config import XXX" which would make such a feature rather hard to do.

The "config.py" file was initially present to offer a some kind of default customization for the environment. What I mean by that is:
- grid2op has some default behaviour, shared for all environment
- config.py offers some basic default configuration, different from the config for each environment
- a call to grid2op.make(...) can (at least in theory, if that is not the case then we can work on it for sure) override any "default" configuration.

# First "solution"

So I am not entirely sure 2 configurations are needed. If you want 2 confs you can store them (as yaml, json or serialization mechanism you want) and do something like this (example with json):

```python
import json
import grid2op
with open("environment/folder/configs/config_1.json", "r", encoding="utf-8") as f:
    conf1 = json.load(f)

with open("environment/folder/configs/config_2.json", "r", encoding="utf-8") as f:
    conf2 = json.load(f)

env_conf1 = grid2op.make(env_name, **conf1)
env_conf2 = grid2op.make(env_name, **conf2)
```

# Second "solution"

Or you can also do an "env with config" function that will "bind" the two above call if that is more convenient for you, for example:

```python
import grid2op
import json
env_name = "environment/folder/"

def make_with_config(env_name, config_name):
    with open(os.path.join(env_name, config_name), "r", encoding="utf-8") as f:
        conf = json.load(f)
    return grid2op.make(env_name, **conf)
```

What could also be added in grid2op is this feature in env.make with either:
[...]

PR #739, opened 2026-02-12
feat: Make env using custom config file path in kwargs

kostialjn:
Specify custom config.py file using kwargs. Related to issue #738 

```python
# Load env using config.py in root directory
env = grid2op.make(root_path)

# Load env using custom config.py anywhere else
env = grid2op.make(root_path, _config_path="path/to/custom/config.py")
```

### DIFF
--- a/grid2op/MakeEnv/MakeFromPath.py
+++ b/grid2op/MakeEnv/MakeFromPath.py
@@ -339,7 +339,10 @@ def make_from_dataset_path(
     _check_kwargs(kwargs)
 
     # Compute and find config file
-    config_path_abs = os.path.abspath(os.path.join(dataset_path_abs, NAME_CONFIG_FILE))
+    if "_config_path" in kwargs:
+        config_path_abs = os.path.abspath(kwargs["_config_path"])
+    else:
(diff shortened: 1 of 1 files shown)


BDonnot:
Hello, 

Is this PR still relevant or have you find a better solution ?

kostialjn:
@BDonnot 

What we did was : 

- Create a fake empty config at env folder root 

```python
# Content of config.py
config = {}
```

- Create several custom config.py (config1.py, config2.py, etc)
- Import the one needed and load environment with it by assigning them through kwargs: 

```python
from .config1 import config
env = grid2op.make(..., **config)
```

The only problem was that some kwargs names are incompatible with some config.py fieds see issue https://github.com/Grid2op/grid2op/issues/740 for instance

BDonnot:
Get it 😊 

This will be fixed in next release.

I will double check every config names to match the kwargs names.

But in the same time and before the release, I wanted to merge anything that could be merged.

This is why I wanted to check whether this makes sense to merge this PR.

BDonnot:
Closing it as i'm not sure it's still relevant today. Let me know if it's still usefull.

PR #746, opened 2026-04-14
Some fixes: copy on write in pandapower and lightsim2grid issue 128

BDonnot:
- [FIXED] copy on write issues in PandaPowerBackend
- [FIXED] a bug causing https://github.com/Grid2op/lightsim2grid/issues/128
- [FIXED] some warnings in the docstrings (escaped character)


BDonnot (commented):
looks good

PR #761, opened 2026-05-31
docs: clarify reward kernel flags

### LINKED ISSUE #737: Reward Kernel Formulation - incorrect in docs
This was discussed and confirmed in [https://discord.com/channels/698080905209577513/702060196222205962/1468594043851374604]

Assume the grid has only 1 redispatchable generator, and max ramp up/down is +5/-5, according to the docs, the action space is [-5, 5]   [https://grid2op.readthedocs.io/en/latest/mdp.html#modeling-sequential-decisions]
 
If the agent outputs an action like [6.0] (this can be the case if the agent is a neural network), in this case, the way the reward kernel behaves, in the language of MDP, doesn't match the notation.

The reward kernel will process and perform "do-nothing action".
But for some reward functions available in Grid2Op, they can give a "-1.0" reward signal because the agent asked for an illegal action.

In the notation of reward kernel, the action "a", is it:
- Case A: the illegal, out-of-action-space [6.0],
- or Case B, the "do-nothing action" (which replaces the [6.0])?

I'm seeing a little of contradicts here:
If it's Case A, that means the reward kernel is processing an action (or action vector) that doesn't belong to the action space of the environment (out of [-5, 5] )? Is that suitable?
If it's Case B, it doesn't make sense, because now it returns -1.0 for a do-nothing action, which won't happen if we use the real do-nothing action at the beginning (it will not be treated as illegal, hence not -1.0). So that means there is something outside the reward kernel that defines the -1.0 illegal point? 

## Possible solution
Reward kernel should be a function that takes also some flags from the environment (like "is_ambiguous", "is_illegal" etc.) which is not in the actual formulation.

So we might have:   

final_action, is_ambiguous, is_illegal = translate_action(action_vector_from_agent)

RewardKernel(s, final_action, is_ambiguous, is_illegal)

final_action usually is the do-nothing action

william-xue:
## Summary

- Clarify how Grid2Op reward classes receive contextual flags in addition to the usual MDP reward-kernel inputs.
- Explain that these flags distinguish the agent-submitted action from the action effectively applied by the environment, e.g. when an out-of-bounds redispatching action is replaced by do-nothing.

## Validation

- `git diff --check`
- `python3` content check for the added reward documentation note

Closes #737

### DIFF
--- a/docs/user/reward.rst
+++ b/docs/user/reward.rst
@@ -31,6 +31,13 @@ Customization of the reward
 In grid2op you can customize the reward function / reward kernel used by your agent. By default, when you create an
 environment a reward has been specified for you by the creator of the environment and you have nothing to do:
 
+.. note::
+    In the mathematical MDP notation, the reward kernel is often written as a function of the state,
+    the next state and the action. In grid2op's implementation, reward classes also receive contextual
+    flags such as `has_error`, `is_illegal` and `is_ambiguous`. These flags make it possible to distinguish
+    the original action submitted by the agent from the action effectively applied by the environment, for
+    example when an out-of-bounds redispatching action is replaced by a do-nothing action.
+
 .. code-block:: python
 
     import grid2op


BDonnot:
Hello,

Thanks again for this fix.

Same (minor) issue as for the your other one. You did not signoff (Developer Certificate of Origin (DCO) ) your commit.

Doing it is really easy in your case, you simply need, locally to type:

> git commit --amend -s
> git push --force

should do the trick.

The first line will simply add a "signed off by <XXX>" in the message commit, which is required to contribute to open source projects hosted by the linux fundation. See https://developercertificate.org/ for more information.

Thanks again and let me know if you encounter any issues

william-xue:
Thanks for pointing this out. I amended the commit locally with a DCO sign-off; pushing the rewritten branch requires force-with-lease, so I will update the PR branch as soon as that guarded push is allowed.

BDonnot (approved):
ok

BDonnot:
Thank you very much for this contribution :-)

